Issue #2, opened 2021-03-24
use setnames and setattr data.table helpers rather than names<- and attr<-


d-callan:
this to update names and attributes by reference rather than making copies of the data tables

asizemore:
Closed with PR #13

PR #13, opened 2021-03-31
Fix/2/setattr: fix for issue #2 by changing to setattr and setnames

asizemore:
# Fix for issue #2 

The fix required two parts:
1. Change assignments `names(dt) <- ...` to use the data.table function `setnames`.
2. Change assignments `attributes(dt) <- ...` to use the data.table function `setattr`.

For part 1, a simple find and replace solution worked well. Note that `setattr` does not accept type "list". Must be a data.table or data.frame.

For part 2, a similar approach would not work, because often we perform a pattern in which we (1) assign attributes of an object to `attr`, (2) do things to the object, and then (3) add all of the attributes in `attr` back to the object. So what we really need is to replace the last step of this pattern with `setattr(dt, attr)`. Alas, `setattr` does not add attributes from a list, so instead `utils.R` now includes a `setAttrFromList` function.

## Testing
The above changes were tested with `devtools::test()`.

After commit 641521e, testing returns the following two new failures:

```Failure (test-histogram.R:125:3): histogram() returns consistent and appropriately formatted json
`outJson` has changed from known value recorded in 'histogramJson.rds'.
1 string mismatch
Backtrace:
 1. testthat::expect_equal_to_reference(outJson, "histogramJson.rds") test-histogram.R:125:2
 2. testthat::expect_known_value(..., update = update)

Failure (test-histogram.R:130:3): histogram() returns consistent and appropriately formatted json
names(jsonList$config) not equal to c(...).
Lengths differ: 7 is not 6
```

The errors are due to the following change in `plots-histogram.R` starting at line 99:

```  
### This change causes 2 new histogram errors!
  # attributes(.pd) <- attr
  setAttrFromList(.pd, attr)
```
The original code replaces _all_ attributes of `.pd` with those listed in `attr`. However, before this step, `.pd` has an attribute, "sorted", that is not also in `attr`. So, after `attributes(.pd) <- attr`, the `.pd` object no longer has "sorted" as an attribute. 

While in this instance, it seems like "sorted" might be useful and me may want to keep it, I am realizing right now as I'm writing that likely we want to be very specific with which attributes exist since they all get written to the json file eventually. So, perhaps it would be best to have `setAttrFromList` first remove all attributes not in `attr`, and then update/add attributes in `attr`. 

Fast forward to commit bfe9d4c...
The above histogram errors are no more. If we want to go this route of replacing the attributes of the input with those in `attr`, I would consider renaming the function to `replaceAttrWithList` or something similar.

_Note:_ This PR is still a draft as I've left extra comments and print statments in for later testing following the outcome of the above discussion.


d-callan:
i could be wrong, but i believe the 'sorted' attr is to do w the keys on the underlying data.table. i dont think we'll need it for the histo class, and am happy to drop it for now and revisit later if we decide at some point that we need to retain the keys in the histo class.

asizemore:
Perfect. I made replacing vs. updating an argument to the `setAttrFromList` function so that we can decide on the fly as we move forward. Or at least it will be easy to change.

d-callan:
couple thoughts:
1. most important thing is to avoid for loops. theyre notoriously slow in R. there is a package called microbenchmark you can use to give you an idea just how slow. i would probably make a helper to pass to lapply or something. putting something together quickly as an ex: removeAttr <- function(attrToRemove, dt) { data.table::setattr(dt, attrToRemove, NULL)} and then the lapply would be something like lapply(attrsToRemoveList, removeAttr, dt) . 
2. could probably use is.data.table rather than check for the class w %in%, but i dont mind either way.
3. we might need to find a better name for our replace arg to make more clear what it does. though im not sure what to call it really, remove isnt much better.

asizemore:
Excellent. Recent commits address all three points. Specifically, new suggest changing 'replace' -> 'removeExtraAttrs'

d-callan (approved):

Issue #7, opened 2021-03-24
determine possible options for value arg in scatter class


d-callan:
currently the options are 'smoothedMean', 'density', and 'raw' . but all options actually return raw data, w smoothed mean and density just returning those data in addition to the raw data. do we want this? or break raw into its own boolean arg? or something else?

d-callan:
After some conversation in the data viz channel I think the options we need are 'raw', 'density', 'smoothedMean' and 'smoothedMeanWithRaw'

asizemore:
Closed with PR #14

PR #14, opened 2021-04-01
Fix for issue #7 adding "smoothedMeanWithRaw" to value argument

asizemore:
# Fix for issue #7 

Originally the options for the scatterplot value argument were "raw", "density", and "smoothedMean". However, both "density" and "smoothedMean" also returned the raw data. Instead, we want the raw data to only be returned when called for explicitly. 

The proposed fix adds a "smoothMeanWithRaw" option for the scatterplot `value` argument, and removes the raw data from being merged with the smoothedMean interval or density when `value='smoothedMean'` or `='density'`. 

## Testing
The above changes created new, but expected failures in the following test:

```
scattergl.dt() returns an appropriately sized data.table 
```
The new data.tables are missing two columns, which is expected because they no longer contain the raw data. 

Once aligned on the fix and implementation, should we add a test or two for the new argument?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,3 @@
+# .gitignore
+
+.DS_Store
--- a/R/plots-scatter.R
+++ b/R/plots-scatter.R
@@ -40,20 +40,28 @@ newScatterPD <- function(.dt = data.table::data.table(),
     interval <- groupSmoothedMean(.pd, x, y, group, panel)
     interval <- interval[, !c('ymin', 'ymax')]
     names(interval) <- c('interval.x', 'interval.y', 'interval.se', group, panel)
+    
+    .pd <- interval
+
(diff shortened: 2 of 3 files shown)


d-callan:
looks good. i agree we should update the existing tests to get them passing again and add tests for the new option. we also need to make sure we update the documentation.

asizemore:
Great! Tests added and `scattergl` documentation updated. Note the above changes have altered the ordering of `names(dt)` (see changes to `test-scattergl.R`).

d-callan:
only little thing i notice is that the documentation looks to be updated for only one of the two exposed scattergl fxns

asizemore:
Good catch! My mistake -- fixed in the latest commit.

d-callan (approved):

Issue #3, opened 2021-03-24
update x,y,group to independent,dependent and overlay


d-callan:

d-callan:
i still think we need to clarify names.. but depending on how conversations around client side swapping of axes and gradient colors scales etc go. so maybe hold on this for now

asizemore:
let's remove z entirely. It'll be overlay now

asizemore:
decisions: call group overlay, panel is panel still (not subplot), facet still facet, x to independent, y to dependent

PR #19, opened 2021-04-22
Fix/3/varname update: fix for issue #3 by clarifying variable names

asizemore:
# Fix for issue #3 

This fix required modifications throughout the package. In general,
- 'x' -> 'independent'
- 'y' -> 'dependent'
- 'z' -> 'gradient'
- 'group' -> 'overlay'

*Notes:*
- Left 'x' if it was referring to a general object or vector. We could consider replacing 'x' with 'vec' or 'v' as well.
- Current testing has multiple expected failures when comparing to saved data.
- Viewport variables were left unchanged.
- Changed a few vars in `predictdf` to clarify predicted values (we should discuss what is most common here)
- After we agree on all variable names i'll update test data
- Draft PR


asizemore:
too stale. closing

Issue #4, opened 2021-03-24
add arg to group fxns to make noStatsFacet optional


d-callan:
if a group fxn relies on another (ex: groupSmoothedMean finds max groupSize to know what method to use) we dont want to collapse the data w noStatsFacet. but when the plot classes call group fxns they do want to collapse the data to one row per group.

asizemore:
I'm not sure I quite understand -- in the groupSmoothedMean/groupSize example, groupSize currently does not call noStatsFacet. Is the issue that groupSize should have an optional call to noStatsFacet (via an additional arg)? Or that any of the group functions that currently call noStatsFacet (groupMean, groupFences, etc.) should get an additional arg to make their noStatsFacet call optional, regardless of if these functions are currently called by another group function?

asizemore:
Closed with PR #20

asizemore:
Closed with PR #20

PR #20, opened 2021-04-23
Fix/4/no stats facet

asizemore:
# Fix for issue #4 

This PR addresses the issue by
1. Updating `groupSize` to optionally run `noStatsFacet`.
2. Adding the option to run `noStatsFacet` to other group functions.

<br>

**Additionally** this PR renames `noStatsFacet` to `collapseByGroup` and calls the new argument passed to group functions `collapse` (set to TRUE by default).
<br>

*Notes:* 
- The functions `groupDensity` and `groupSmoothedMean` did not receive the optional `collapse` argument because they already perform their own collapsing. May be worth revisiting in the future.
- The new name for noStatsFacet is still up for debate.
- Column names in `plots-bar.R` had to be reordered to accommodate now running noStatsFacet before naming instead of after.

### Testing
All expected tests passed. Only failures were the usual two while testing that "groupSmoothedMean() returns consistent results".


d-callan (approved):

d-callan:
if any of these are exported, we should make sure the documentation is up to date as well before merging.

Issue #12, opened 2021-03-29
Box plot to take list of x or y axis vars


d-callan:
This needed for the rel abundance mbio app. Need to think carefully about validation, since all selected terms must measure the same thing. This likely means subclassing the boxplot class.

asizemore:
Closed with PR #39

Issue #5, opened 2021-03-24
improve api for providing var id to plot ref map


d-callan:
clunky i think to expect two data.tables as args to the plotting functions. the second is used for mapping variables to their plot references (x,y,overlay,facet) and data types etc. we may be able to find a better way to do this, both for the exposed helper functions and the lists used by the underlying s3 classes.

d-callan:
i think were going to live with it

PR #21, opened 2021-04-27
Fix/12/box listvars: allow box plot to accept specified xAxis variables

asizemore:
# Fix for issue #12 

Goal: Allow allow one to specify an arbitrary list of numeric columns in the data to be used as x-axis values (i.e. one box per column).

The main update with this code is a function called `reshapeByListedVars`, which takes in the data table, an appropriate list of variables, and expected variable type, and melts the data from wide to long based on these variables. The function itself really just wraps data.table::melt, but adds extra checks to ensure we're using appropriate variables.

Currently, we have a passable solution but I would like to improve it. Before , we should discuss the following:
1. How should multiple variables get encoded into `map`? Worth thinking about in tandem with issue #5. Currently we let "xAxisVariable" elements in `map` be strings that get split by a specified delimiter. Alternatively, we could let "plotRef" hold multiple "xAxisVariable"s, and we can simply check to see if there is more than one "xAxisVariable". Currently we use the former because it feels less likely for the variable ordering to get mixed up. Though the latter would remove the need for a delimiter so that's very attractive...
2. When and how to set the `varOrder` attribute? Ideally we set this inside `newBoxPD`. I am considering the following
     - Get `varOrder` from the original `map` passed to `box.dt`. Easy to do but we have to alter `map` before it is sent to `newBoxPD` and this alteration removes the variable ordering information. So, we would have to send the info to `newBoxPD` through a new function argument.
     - Get `varOrder` from the levels of the reshaped data.table. If we use `data.table::melt`, then we have the option to set the new `variable` column in the output data.table to be factors with the appropriate ordered levels. If we go this route, then when we send data to `newBoxPD`, inside that function we can use `levels()` to extract the ordering information, and set the attribute right there in `newBoxPD`. That's great, but it means we are relying on data columns being factors in certain cases and not others, and we don't have factors elsewhere so this option feels not so robust. <br>
     
For now I'll choose the first of passing the variable order information as a new argument to `newBoxPD`.

3. The "custom input variable" language could use some help. Picked variables? 
4. When Dave returns, I'll ping him about the new attribute coming his way.

Considering renaming `reshapeByListedVars` to `reshapeByVarList`. Hmm.

## Testing
Added new unit tests for `box.dt` and for `reshapeByListedVars` (in utils.R).

## Profiling
todo.

### DIFF
--- a/R/plots-box.R
+++ b/R/plots-box.R
@@ -123,6 +123,7 @@ box.dt <- function(data, map, points = c('outliers', 'all', 'none'), mean = c(FA
   facetVariable2 = list('variableId' = NULL,
                         'entityId' = NULL,
                         'dataType' = NULL)
+  varOrder <- NULL
 
   if (!'data.table' %in% class(data)) {
     data <- data.table::as.data.table(data)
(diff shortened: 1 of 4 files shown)


asizemore:
Currently on hold until we have a better idea of how many plots might use this idea and in what contexts (overlay, facet, etc).

asizemore:
See [PR](https://github.com/VEuPathDB/plot.data/pull/39). Closing this draft.

Issue #25, opened 2021-05-11
write scripts that profile plot.data package


asizemore:

d-callan:
we can probably keep these scripts in the package itself. we can create a top level directory where its name starts w '.' and it should be ignored when the package is built and installed.

asizemore:
Closed with PR #32

PR #32, opened 2021-05-17
Feature/25/profiling

asizemore:
# Profile plot.data

Addresses issue #25. In the spirit of testthat, this PR adds a collection of scripts in `./.dev` folder that each perform a `microbenchmark` tests on functions within a particular context. Quick outputs of time differentials from the previous logged run are printed to help pinpoint problem areas.

The `.dev` folder new contents:
- `microbench-*.R` runs `microbenchmark` on relevant functions within the context. These files can be run as standalone scripts or all together from the `run-all-microbench.R` script.
- `run-all-microbench.R` runs all `microbench-*.R` files.
- `helpers-microbenchmark.R` contains functions that support other scripts in the folder.
- `benchmarks.rds` contains the summarized results from the last saved run. These data are used for comparison to the current run.

Two use cases drove the organization of the code. First, optimizing a single function (such as `histogram.dt`). 
For this case, we only want to check how changes affect relevant code. So, any `microbench-*.R` could be run with `overwrite=F` repeatedly to understand how changes to the code affect run time. After optimization is complete, we could similarly update only the relevant context with `overwrite=T`. 
Second, ensuring new code does not cause slowdowns across the package. Here we would run `run-all-microbenchmark.R` which will loop through all benchmarking files and print the outputs. When each script is run, the `overwrite` parameter in each individual script will take the value of `allOverwrite` parameter defined in this script.


asizemore:
At this point I just grabbed a few function calls from the testthat scripts to profile. Happy to add a bunch more if that would be helpful!

d-callan:
this is a fun and potentially useful idea. we should double check it doesnt affect the package building/ break rserve. but otherwise i dont see any harm in merging, and seeing what this grows into.

d-callan:
should consider something similar for Rprof and summaryRprof functions maybe. but fine to merge, and maybe ill expand on this when i get to profiling umsp (next week maybe)

PR #34, opened 2021-05-18
Feature/28/bar to take proportion

asizemore:
# Feature: Bar returns proportion

In response to issue #28, this PR adds an additional option to the bar plot. Specifically, the `value` argument for `bar.dt` and `bar` now accepts one of `c('count', 'identity', 'proportion')`. 

The majority of the heavy lifting is done by the new function `groupProportion` which follows the same structure as `groupSize` but returns proportions per group instead of counts.

Added new tests in `test-bar.R` and `test-group.R`.


d-callan (approved):
i like it. think we should merge it.

Issue #26, opened 2021-05-12
incompleteCases needs to become a table 


d-callan:
we want to have a table listing all variables playing into the plot, and the number of records missing data for each. this to replace the incompleteCases attribute.

d-callan:
rather than replace, i think this could be in addition to the incompleteCases attr as the existing one i think is useful for the subset tab. could possibly be named better as well? getJSON helper fxn and the plotdata class both need updating at a min. and when this is done, the api in the data service needs updating as well before we can deploy this change.

d-callan:
just spoke w steve and dave, theyre happy to do any needed math in the client if we pass them the number of complete records per variable

PR #36, opened 2021-05-19
Fix/26/incomplete cases table

asizemore:
# Return complete cases as a table

Fix for issue #26 .
This PR creates a new table `completeCases` that contains the number of non-NA records for each variable. The number of non-missing values is calculated with `stats::complete.cases`. The `completeCases` table then becomes an attribute of a plot.data object. Finally, the `getJSON` function was updated to handle the new table.

## Testing
New tests for missing data were added to each plot context. Additionally, json tests that did not involve comparing to saved data were updated. No new failing tests.

## Additional Notes
- The new tests added 9 warnings, which all come from chi squared warnings in the mosaic context.
- Tests for histogram that compare json to a saved .rds are currently commented out. Once we are happy with the completeCases json representation I'll uncomment and update the .rds files.


asizemore:
Added new function `addVariableDetails` which basically wraps `makeVariableDetails` and allows easier application for the sampleSizeTable and completeCasesTable.

asizemore:
Awesome. Updated in `utils-json.R`. Additionally I updated the saved histogram json rds files in the testing directory so that now all tests are operational and passing!

d-callan:
decided to update the api some. i wanted the output json to conform the the VariableSpec obj in the data service rather than the StrataVariableDetails obj. interested to hear your thoughts

PR #38, opened 2021-05-27
Feature/33/box returns nonparametric statistical test

asizemore:
# Feature: box returns results of nonparametric test

_Addresses issue #33_ 

This PR adds two new functions to `utils-stats.R`
- `nonparametricTest` takes one numeric vector of values and one character vector of levels that break values into groups. Returns results of either the `kruskal.test` or `wilcox.test` based on the number of unique levels.
- `nonparametricByGroup` runs `nonparametricTest` on the `data` input after splitting the data by `byCols`. 

and additionally adds one argument to relevant `box` functions:
- `computeStats` boolean to determine if above stats should be computed.

The above assume we want to run tests on the following groups:
- if no strata variables, run test on y values grouped by x values.
- If facets/panels but no overlay, run test on y values grouped by x values split by panel (1 test per panel)
- If overlay defined, run test on y values grouped by overlay values split by x values and panel (1 test per x value per panel if defined)

## Testing
Currently incomplete. Tests written in `test-box.R` but not yet in `test-stats.R`. Eventually will need json tests as well. The functions themselves could use a few more checks to ensure the inputs are good (noted in the code). Will update all of these after initial review!
Additionally, the stats tests will err if there are not checks to ensure a proper number of values exist within each group. Handling these errors gracefully is still on my todo list.

# Questions
- What all should we return from the test? Currently returning 'statistic', 'p.value', 'parameter', 'method'. 
- The returned 'statsTable' (poorly named) could likely use some updating based on converting to json/matching reasonable endpiont. I started down this road but then realized we already have mosaic returning `statsTable`. Should we have the box stats and mosaic stats (and any other plot stats) be returned in their own individually named tables, or create a unified statsTable structure so that we can use just the one for all plots? Currently we probably don't have a super strong need since this only involves two plots, but in the future, histogram or violin could gain a stats table. 
- Lots of the naming of functions, column names, etc could use help :)
- We previously talked about returning stats for a box always and then the client can decide if they should display them or not. However, while working on this I realized that sometimes running these tests doesn't make sense based on the inputs. For example, if I have chosen my boxes to have xvar = top 10 taxa and overlay=null, these stats don't make sense. This thinking was the reason behind the `computeStats` parameter addition. Alternatively, we could remove this parameter, always compute and send stats, but have the data service decide if they should be ignored based on the inputs?

Interested to hear your thoughts!!


d-callan:
1. i think the 4 values your returning are at the very least a good place to start.
2. i think, at least for now, its okay to just call its 'statsTable' and not worry that the 2x2 statsTable has a different format. we can talk to the ui guys and see what they like the response to look like later, and let that guide any future adjustment.
3. if any place is going to programmatically decide to run stats or not, it should be plot.data.. we just need to be careful to make clear to the user what we did.
4. smoothedMeans have a similar problem, where sometimes the data just isnt there and when we try to generate one for a group it errs. the strategy, for now, is to just pass those errors along. think we should do something similar here.

d-callan:
also you should be able to specify the boolean type in the constructor like `logical()` and decide a default in the helper like `computeStats = c(TRUE, FALSE)` where the first value will be the default. then inside the helper can use match.arg to test that computeStats is a valid value.

d-callan:
the only other thing ill say is that we should confirm w dan hes happy w the rules for when and how to calculate these stats, and then ask dan and clinepi if they have any interest in them for the pass app.

asizemore:
Updates:
- got the okay from Dan on rules for computing stats
- errors from the two tests are captured in the statsTable
- testing improved in box, utils contexts

d-callan:
think this is really close. probably box class needs a helper to grab out its statsTable, and the tests on the stats should maybe check that the types dont change when there is an error vs not. ex: is pvalue numeric in both cases? otherwise, looks good. 
ill add one minor comment, which is that data frames are implemented as lists, so to add a new col w a name like 'newCol' to some data frame 'df' you can do something like 'df$newCol<-1'

asizemore:
Excellent. Moved `statsTable` helper to `class-plotdata.R` and added testing for consistent types on err.

d-callan (approved):

PR #39, opened 2021-06-03
Accepting multiple vars as one

asizemore:
A new approach to issue #12 .

A few motivating thoughts:
- Multiple-vars-as-one could arrive on the x axis (box plot), y axis (heatmap), or possibly "facets" (one panel per var) if we allow.
- Based on both the plot element and the plot type, we may want to do something different with the vars. For example, a box receiving multiple vars for x then sends their values to the y axis. In a heatmap, multiple vars for x then means their values are handled in the z/gradient. I'm thinking about this as "variable flow" right now, for lack of a better phrase.
- Would we ever allow multiple lists of variables? Assuming no, for a long while at least.

Because of the need to be flexible, the goal was to create a small bit of code that could be added to the `{plot-type}.dt` function that simply reshaped the data into what the rest of the function would expect. This approach has two parts:
1. Reshaping the data.table. Done with `data.table::melt`. Kept at the top level for now for sake of readability.
2. Updating the `map` data frame to appropriately represent the melted data. The function `remapVariableList` handles this job, and also takes arguments that specify "variable flow".

The above works pretty well so far. One piece that now I think becomes a separate issue is how/where/when to handle variable order. If we assume that we receive this list of variables in order (or values of any categorical var), is sending this order off to the ds as an attribute of the pd object enough? Or if plot.data needs to preserve the data order or more generally handle orderings, I can make a separate issue for this. 

**NOTE** Draft pull request. Tests not written. Checks not in place. Left comments about appropriate checks in the code to do after we're happy with the basic implementation. Also currently only implemented for boxplot.


d-callan:
i think this is the right direction. i think scatter needs some logic for this too, for the overlay var. and maybe we'll leave it w those two for now, unless mbio phase 1 needs it other places.

d-callan:
order is an interesting question.. i think its going to have to be a separate issue, bc possibly this will end up the clients problem rather than ours. but it needs discussing w a broader group, particularly steve and dave.

d-callan:
so i wonder if listVarPlotRef and newVarPlotRef are ever different?
and i think the var constraints by number of max values needs expanding to match what weve set in the data service. if its a facet var i guess we can go up to 25 is what we said, w a single facet. and an overlay var can go up to 8 since thats how many unique colors blind colors we have.

asizemore:
Good catch. Right now listVarPlotRef, newVarPlotRef, and meltedVarPlotRef are all the same. I think there's a small possibility that in the future these could be different, but for now definitely all the same. 
Also updated the `validateListVar` function to align with the data service constraints. Additionally with the relative abundance app in mind i added a constraint for the x axis.

d-callan (approved):

Issue #27, opened 2021-05-12
gradient color scales for scatter plot


d-callan:
think weve decided that we need to be able to turn off trend lines if the overlay var is continuous. that will mean slightly complicating the validation of arguments, and will require passing data shape through from the data service first.

asizemore:
Closed with PR #43

PR #43, opened 2021-06-10
Feature/27/prepare scatter for gradient colorscale

asizemore:
Addresses issue #27 

This PR allows the overlay variable passed to `scattergl.dt` to be continuous. 

## Rules for continuous overlay variables:
- No trend lines. `scattergl.dt` will err if passed a continuous overlay var and `value != 'raw`. 
- Continuous overlay vars operate as one group. In other words, there is one row in the resulting data.table per panel, not per overlay var and panel (as is the case otherwise).The overlay var is returned as a list that should have the same length as `'seriesX'` and `'seriesY'`.

## Testing
New tests added to the scattergl context. All relevant tests pass.

## Notes
- Instead of being ordered, `c(group, panel, 'seriesX', 'seriesY') `, the returned pd data.table will have columns `c(panel, 'seriesX', 'seriesY', group)`. I do not expect this to be a problem but can reorder if that is not the case!
- Decided to keep the overlay var validation step in validateScatterPD in case we for some reason are send an overlay var with `dataShape=zebra`, etc.

### DIFF
--- a/R/class-plotdata-scatter.R
+++ b/R/class-plotdata-scatter.R
@@ -128,10 +128,11 @@ validateScatterPD <- function(.scatter) {
 #' 'smoothedMeanY' and 'smoothedMeanSE' specify the x, y and 
 #' standard error respectively of the smoothed conditional mean 
 #' for the group. Columns 'densityX' and 'densityY' contain the 
-#' calculated kernel density estimates.
+#' calculated kernel density estimates. Column 'seriesGradientColorscale'
+#' contains values to be used with a gradient colorscale when plotting.
 #' @param data data.frame to make plot-ready data for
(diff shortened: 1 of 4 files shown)


d-callan:
it occurs to me that probably we want to call this thing something like 'seriesGradientColorscale' or similar. and then in utils-json.R when we write out the variable details, include a check for a continuous overlay variable and write that out similarly to the x and y vars. it does allow for the overlay var details to be in two different places in the response depending on its data shape, but i think it might be ok.

asizemore:
latest updates include:

- renamed column to 'seriesGradientColorscale'
- `addStrataVariableDetails` ignores overlay if overlay$dataShape is continuous
- if overlay$dataShape is continuous, overlay written to 'config'
- reorganized a bit of `getJSON` so that in config the order will be x, y, z, overlay instead of overlay, x, y, z.
- added tests for appropriately formatted json

d-callan (approved):

asizemore:
Closes #27

Issue #42, opened 2021-06-10
write helper for validating function arguments


asizemore:
Should wrap `match.arg` or `%in%` when appropriate.

PR #44, opened 2021-06-14
Feature/42/validate arguments

asizemore:
_Addresses issue #42 _

This PR adds a new function `matchArg` that replaces `match.arg` and other means of ensuring the argument of a function matches an approved value. The `matchArg` function was developed from the original `match.arg` function.

# Notes
- `match.arg` uses partial matching. We do not need partial matching so use `stringi::stri_detect_regex` to match the entire word.
- We will assume that any argument should be one value. This assumption has consequences when assigning default values, and how we match the argument to the allowed values.
- `matchArg` works like `match.arg` in the following ways: When `choices` is not passed to the function, we determine `choices` using the parent function. If `arg` was not specified when calling the parent function, it will be set to the first value of `choices`.

# Testing
No new tests implemented. No new warnings nor errors observed.

# To Dos
If we are happy with the implementation, I'll add 1) function documentation and 2) stringi::stri_detect_regex to NAMESPACE for importing.


d-callan (approved):
sorry for the late response. i think this looks good, ill resolve conflicts and merge.

d-callan:
i need it for something currrently, so ill finish up the to-dos along the way

PR #50, opened 2021-07-01
Feature 45 bar proportion with overlay

asizemore:
_Fixes issue #45_ 

This PR adds new functionality to `groupProportion` that allows it to calculate proportion differently based on if the bars are stacked or grouped.

## Major updates:
- `groupProportion` has a new argument `barmode` that determines the denominator of the proportion calculation. Specifically, if `barmode='group'`, then we calculate (x of that value per group)/(all of that group). When `barmode='stack'`, we calculate (x of that value per group)/(all x of that value).
- The bar plot functions now accept a new argument `barmode`, that is passed straight to `groupProportion`.

## Testing
New tests added. No new failures.

## Notes
- In draft status currently to ensure I'm not going crazy in the proportion calculations. After a sanity check, will apply same logic for histogram. 
- Confirmed with Bob, Dave that "barmode", "group", "overlay",  and "stack" are the right terms.


d-callan:
looking at it now, think its all x of a given value by panel. so byCols = c(x, panel) rather than byCols = c(x)

asizemore:
Updates for `binProportion`

If `barmode = "overlay"`, denominator is all values in the group per panel.
If `barmode = "stack"`, denominator is all values per panel.

Tests for `binProportion` and `histogram.dt` added.

d-callan:
tbh the proportional stacked histo seems strange to me as an idea.. but i suppose we should keep it consistent w bar and make the denom (x,panel) rather than just panel ?

asizemore:
Great. binProportion updated by setting the denominator to c('binLabel', panel). 
@myself, when time allows and if we all agree with this implementation strategy, probably worth rewriting to match the groupProprtion function.

d-callan (approved):

d-callan:
i say we merge, and see what people think of it. ds api will need updating too, but ill sort that out some time early next week.

d-callan:
merged. will make sure tests and documentation are up to date. trying to release a new version of plot.data in the next day or two.

PR #56, opened 2021-07-22
Refactor listVars for mbio custom taxa picker

asizemore:
_Addresses issue #54_

# Implementation details
listVariables are sent into plot.dt functions as repeated entries in the map, along with a required `listVarPlotRef` arg that specifies which plot ref is considered the listvar. plot.dt functions validate plot-type-specific rules for listVars, and then create 'listVarDetails' that it eventually sends to `newPlotData`. The `listVarDetails` object contains details about the listVar itself, such as the eventual display name, and of the inferred variable. The inferred variable is specified in the plot.dt function and then assigned to the new var in `newPlotData`. Finally, `newPlotData` melts the data table so that the listVar is now a categorical variable and the inferred variable exists as a column.

## Rules for listVars
- All participating variables must have the same entity, type, and shape (see `validateListVar`)
- Assume that the inferred variable goes on the y axis. We have the flexibility to change it later, if necessary
- If the listVar is assigned to either overlay or facet, `evilMode` must be `F`.
- There is a maximum number of variables allowed to participate in a listVar based on the associated plotRef.
- Scatter can have listVar as overlay or facet1
- Box can have a listVar as x or facet1
- The existing rule that you cant select the same var for two different plot elements should apply to the parent term for terms annotated as listVar (from Danielle).

## Notes and considerations
- The `listVarPlotRef` is to handle the case where only one variable participates in a listVar
- The compute service will send us at least display labels for these listVars, which gave rise to the other two new args
- The order of the listVars is now captured in the `listVariableDetails` json output, part of the plot config. Also looks like there will be some way to [reorder values in the front](https://github.com/VEuPathDB/web-eda/issues/195) end based on a given list so we should be okay.
- The box `statsTable` didn't make sense so opened a new issue.
- Currently, `listVarPlotRef` and the varId for the new list var are the same. They're currently split in case the CS shows us we need to send the listVar varId info to plot.data as well.
- If listVar is a facet1 and there is only one variable in the listVar, we still get a plot with one "facet". We do not collapse to remove any traces of the faceting, even though there is only one panel.
- we have a new `validateMap` function. We could add additional validation like does every var have a dataShape and dataType, etc.

## To Do DS Updates
- Need make a map with duplicated plotRefs
- Pass `listVarPlotRef` and associated args to box and scatter
- Receive a new `listVarDetails` object in the plot config response.

## Testing
- Added new tests for scatter, box. 
- All previously passing tests pass in plot.data.
- Integration tests caused failures, so I moved new box and scatter arguments to the end of the arg list and found some json I needed to fix. 

# Previous comments and descriptions
## To Dos

- [x] Restructure using box x
- [x] Add validation
- [x] Address json (made issue for statsTable)
- [x] Implement for box facet
- [x] Implement for scatter
- [x] Handle case of only one variable chosen in the listvar. Need another argument to notice this because currently we just look for duplicates in map
- [x] Add args to take listvar, listvalue names, displaynames
- [x] allow facet2 to be specified if facet1 is a listVar
- [x] Add tests
- [x] Clean code + document new args

## Details
Recall when we have a listVar, we send in one (ex x-axis) variable and end up with two "internal" variables: the meltedVariable (ex: x ticks) and meltedValue (ex: y values). 

So far, this PR adjusts the json output for a listVar so that it would be returned as a list of variableDetails that correspond to the original variables. Additionally, checked that when listVar is on the x-axis, the boxplot output dt complies with the ds api.

Open questions surround how do we want to handle both the meltedVariable and meltedValue in plot outputs? Specifically, statsTable, completeCasesTable, sampleSizeTable, and the dt when listVar is facet.

For example, for listVar acting as x-axis in boxplot, completeCasesTable currently would output

| VariableDetails      | completeCases|
| ----------- | ----------- |
| meltedVariable     | num       |
| meltedValue      | num       |
| overlayVariable   | num        |

For this table, we could break up meltedVariable into its respective original varIds (taxon1, taxon2, taxon3) and delete the meltedValue row.

Other times we probably just want to figure out how we want to display the listVar in these tables. For example in the statsTable we currently output
| meltedVariable      | statistics |
| ----------- | ----------- |
| taxon1      | num       |
| taxon2   | num        |

Here we'll need to assign some column name to the table to replace "meltedVariable". This is the same situation for sampleSizeTable

For now since we're only allowing listVar in relative abundance for phase 1, we could assume (in the DS) that meltedVariable name will always be "taxa" and meltedValue would be "abundance"? 

### Details about implementation
This PR changes the listVar slightly to call the meltedVariable `reduce(listVarIdOrder, function(x,y,sep="__") {paste(x,y,sep=sep)})`. This way the var can internally act like any other variable and it's only when we're done creating `.box` that we go back to handle listVars. Specifically, I'm hoping to keep listVars out of `newBoxPD` and handle them all in `box.dt` and when writing json since listVar manipulation deals with getting data into a plot ready format, and not so much with making a boxplot.


d-callan:
complete cases table should probably show the originally selected vars. the others can maybe say 'selected variables' or some similarly vague thing?

R/class-plotdata-box.R
```diff
@@ -209,13 +208,16 @@ box.dt <- function(data, map, points = c('outliers', 'all', 'none'), mean = c(FA
 
     # Record variable order
     listVarIdOrder <- map$id[map$plotRef == listVarPlotRef]
+    #### Sep cannot include . (see plotRefMapToList)
+    meltedVariableName <- as.character(purrr::reduce(listVarIdOrder, function(x,y,sep="__") {paste(x,y,sep=sep)}))
```
d-callan, line 212: paste(listVarIdOrder), collapse='__')

d-callan:
had a thought.
think box.dt and scatter.dt should do some checks that if there are duplicated plotRefs, they are valid ones. but the rest of the logic should probably be moved to plot.data class. that way we can find completeCases based on the individual taxa vars easily, create attributes with meaningful names/ values still, etc. so an xAxisVariable attribute of a plot could potentially be a list then. and for things like the statsTable the header could be 'xAxisVariable' or whatever instead of 'meltedVariable' maybe?

asizemore:
Wrote a very basic draft skeleton of what having the melting step in the plot.data class might look like (see new changes in box and plotdata classes). So far the changes address the case of having x-var as list var only. The box class would handle validation, as you suggested.

Advantages: the attribute tables work out great. And it definitely makes more sense to have complete cases per taxon chosen in the listvar. This is a huge help.

Disadvantages: Either we always decide that if listVar is x-axis, then its values go on the y axis (etc for any plotRef), or we need to have the plot type tell plot.data class where the values of the y-axis should go. Basically we either hard code the list var routes in plot.data class or add another argument to the constructor. 

One solution to the above would be to have the plot type send in a valueVariable with a variableId of 'yAxisVariable' and then it could be assigned more easily. I'll try implementing that next.

asizemore:
The first commit above now adds a little logic in the box class, and a little in the plot.data class. It's more flexible as we expand listvar to other plotRefs and types, and doesn't complicate the logic in plot.data constructor too much.
The second adds my test data to `test-box.R`.

Note we'll still have a pain point in writing to json, because things like the sampleSizeTable require that the xAxisVar can be converted to a column name in that table but it won't be able to in the listvar case. 
Not trying to be a downer, just hoping to highlight pain points so that we can be strategic about where/when that pain will be handled.

d-callan:
Yea. I'm still inclined to thinking this is the better way to handle them. I'm fine adding a new arg to the plot.data class constructor. And think we can sort out the json. The plot obj should know, by it's attributes, if it has a listVar. Writing json will have to learn how to check for them and handle them I suppose.

asizemore:
Nearing the end!! Quick naming question. We have two concepts to describe, basically the list variable and the values of that variable that will become the y axis. 

Some options:
- 'listVar' and 'listValue' . Current. Decent but sometimes we call the whole concept a 'listVar' so can get confusing when then it needs to point to exactly the x axis or something)
- 'meltedVar' and 'meltedValue'. Has the pro of being more commonly used in R and tidyverse.
- 'repeatedVar' and 'inferredVar'. Looses the variable/value distinction but could be more clear to a non R user how these get used in a plot.

Any of the above stick out to you as amazing? Or do you have other ideas?

d-callan (commented):
big step in the right direction, definitely.

d-callan:
i dont think i have a strong preference currrently for naming these things btw. just that whatever we decide we are internally consistent, so if we talk about a listVar in some meeting and then someone from the group goes to read this code, they might know what it is.

R/class-plotdata-box.R
```diff
@@ -183,45 +200,73 @@ box.dt <- function(data, map, points = c('outliers', 'all', 'none'), mean = c(FA
...
+                         'listVarPlotRef' = listVarPlotRef,
+                         'listVarDisplayLabel' = listVarDisplayLabel)
+
+  if (!is.null(listVarPlotRef)) {
+    if (identical(listVarPlotRef, 'xAxisVariable')) {
+      
+      # Ensure all variables are numbers
+      if (!all(xAxisVariable$dataType == 'NUMBER')){
+        stop("listVar error: All overlay vars must be of type NUMBER.")
+      }
```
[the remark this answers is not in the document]
asizemore: in the current scoped uses for listvars, yes. If we let them work with other kinds of data, then I think it would need to be plot-specific. For example, if we let a user choose a listvar for the bar x axis, then we'd need to ensure the dataShape was categorical.
I'll move to plot.data class until we hit that scenario, though. It'll be easy to split up if we come to that. I'll aslo fix that error message :)

R/class-plotdata.R
```diff
@@ -89,7 +98,80 @@ newPlotdata <- function(.dt = data.table(),
...
+    y <- toColNameOrNull(yAxisVariable)
+    yType <- emptyStringToNull(as.character(yAxisVariable$dataType))
+    yShape <- emptyStringToNull(as.character(yAxisVariable$dataShape))
+    .dt[[y]] <- updateType(.dt[[y]], yType, yShape) 
+
+    data.table::setcolorder(.dt, c(x, y, z, group, panel))
+
+  }
+  
+  # If overlay is continuous and NOT a listvar, it does not contribute to final groups
```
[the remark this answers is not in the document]
asizemore: I'd love to do that. The reason the reshaping comes at the end is so that the supporting tables (ex. complete cases) work out nicely. But I'll have a go at seeing if we really need to updateType at the beginning or if that can get moved further down.

R/class-plotdata-scatter.R
```diff
@@ -223,21 +218,42 @@ scattergl.dt <- function(data,
...
+  # Handle listVars
+  listVarDetails <- list('inferredVariable' = NULL,
+                         'inferredVarPlotRef' = 'yAxisVariable',
+                         'listVarPlotRef' = listVarPlotRef,
+                         'listVarDisplayLabel' = listVarDisplayLabel)
+  if (!is.null(listVarPlotRef)) {
+    if (identical(listVarPlotRef, 'overlayVariable')) { inferredVarEntityId <- unique(overlayVariable$entityId)
+    } else if (identical(listVarPlotRef, 'facetVariable1')) { inferredVarEntityId <- unique(facetVariable1$entityId)
+    } else if (identical(listVarPlotRef, 'facetVariable2')) { inferredVarEntityId <- unique(facetVariable2$entityId)
+    } else { stop('listVar error: listVarPlotRef must be either xAxisVariable or facetVariable1 for box.')
```
d-callan, line 247: small thing. but that message doesnt match the tested conditions any longer

R/utils.R
```diff
@@ -471,37 +473,66 @@ remapListVar <- function(map, listVarPlotRef, newValuePlotRef, newVarId = 'melte
...
+
+    # Check we do not have too many repeated variables.
+    nVars <- sum(duplicated(map$plotRef))
+
+    if (repeatedPlotRef == 'xAxisVariable' & nVars > 10) {
+      stop("Too many values specified with listVar: maximum number of x axis values is 10.")
+    } else if (repeatedPlotRef == 'overlayVariable' & nVars > 8) {
+      stop("Too many values specified with listVar: maximum number of overlay values is 8.")
+    } else if (repeatedPlotRef == 'facetVariable1' & nVars > 10) {
+      stop("Too many values specified with listVar: maximum number of panels allowed is 10.")
```
d-callan, line 521: should double check the rules used in the data service for facets.. i think we said the total number of panels could be larger than 10, and we dont limit based on individual facet vars. also this conditions tests on a specific facet, and then the message talks about panels which are the interactions between both facets.

d-callan (commented):
i like the recent changes. think im ok just updating the panel col, rather than the individual facet var columns. i have a vague, and possibly unfounded, concern about receiving a factor for one or both facet vars. but i think if we test for that and it turns out to be a problem, then probably the function that returns the panel col given the two facets can handle that case..

asizemore:
updated error messages, added logging, added number of panels check, and added tests to ensure if we accidentally get factors as facet variables everything still works.

R/class-plotdata-scatter.R
```diff
@@ -244,7 +244,7 @@ scattergl.dt <- function(data,
     if (identical(listVarPlotRef, 'overlayVariable')) { inferredVarEntityId <- unique(overlayVariable$entityId)
     } else if (identical(listVarPlotRef, 'facetVariable1')) { inferredVarEntityId <- unique(facetVariable1$entityId)
     } else if (identical(listVarPlotRef, 'facetVariable2')) { inferredVarEntityId <- unique(facetVariable2$entityId)
-    } else { stop('listVar error: listVarPlotRef must be either xAxisVariable or facetVariable1 for box.')
+    } else { stop('listVar error: listVarPlotRef must be either overlayVariable, acetVariable1, or facetVariable2 for scatter.')
```
d-callan, line 248: typo in facetVariable1

R/utils.R
```diff
@@ -517,8 +517,10 @@ validateMap <- function(map) {
       stop("Too many values specified with listVar: maximum number of x axis values is 10.")
     } else if (repeatedPlotRef == 'overlayVariable' & nVars > 8) {
       stop("Too many values specified with listVar: maximum number of overlay values is 8.")
-    } else if (repeatedPlotRef == 'facetVariable1' & nVars > 10) {
-      stop("Too many values specified with listVar: maximum number of panels allowed is 10.")
+    } else if (repeatedPlotRef == 'facetVariable1' & nVars > 25) {
+      stop("Too many values specified with listVar: maximum number of facet1 values allowed is 25.")
+    } else if (repeatedPlotRef == 'facetVariable2' & nVars > 25) {
+      stop("Too many values specified with listVar: maximum number of facet2 values allowed is 25.")
```
d-callan, line 523: do we/ can we test somewhere the interaction of the two facets results in fewer than 25 unique values?

d-callan (commented):
were so close!

R/utils.R
d-callan: i found it. ignore.

d-callan (approved):
all i can see is a typo, so ill APPROVE!

asizemore:
Yay!!! Until we meet again in the data service, listVars!

PR #63, opened 2021-07-29
set timezone on package load

asizemore:
Example for setting vars on package load.


asizemore:
Decided on the .env file approach since it does not quietly alter the user's environment. 

*To use*
```
library(dotenv)
load_dot_env(file=".dev/.env")
```

Note since this package is for dev only, it is not included in the namspace file.

PR #66, opened 2021-07-30
draft implementations for upset calculations

asizemore:
To match Bob's [PR](https://github.com/VEuPathDB/web-components/pull/168) in web-components, this PR adds computations necessary for upset plot implementation. 

See `upset-prep.R` for the latest example.

The `UpSetJS` component takes `sets` and `combinations`. The former contains data for the highest level sets, so the total number of values that have data for 'A', 'B', etc. The latter contains information on the cardinality of the intersections of these sets. 

The current example takes a data table and returns a list of combinations that could be written to json and sent to UpSetJS. 

# Open
- `elems` should write to an empty array, not an empty object
- still need to implement a method to create the `sets` object, but should be straightforward (can use complete cases table)
- From upset.js, the "sets" value is set to `sets: new Set(containedSetsObjects)`. Not sure what to do about that just yet.
- If it becomes impractical to generate the final `sets` and `combinations` to pass to the component, we can still use this code to generate the data needed for `extractFromExpression`. 
- Still of course needs thought around where these calculations would go, structure, etc., this draft pr is more just a proof of principle to help us evaluate upset.js


asizemore:
After more thought, `extractFromExpression` does enough formatting that maybe the R should just send the sets and cardinalities, and not worry about the nice formatting

Issue #61, opened 2021-07-29
Pass along var display labels when provided


d-callan:
Think we need an optional col in map for display names of vars. If present, pass them along as part of var details in the json response. Needed for compute vars that supply their own display labels.

asizemore:
Resolved by #67

PR #67, opened 2021-08-02
Feature/61/pass display labels

asizemore:
*Fix for issue #61*

This PR adds an optional `displayLabel` column in the `map` data.frame. If there is a `displayLabel` column, the entries may be '' (indicating no display label) or a string.

The above required minor updates to plot.data classes, `plotRefMapToList`,  `remapListVar`, and many functions in `utils-json.R`.

# Notes
- If no displayLabel is given, will be assigned to NULL
- displayLabels are only written to json if they exist
- will require update to VariableSpec?
- Handling listVars isn't great. Right now we have the option of passing displayLabels for the meltedVar and meltedValue vars, but we currently don't have a way to send such values through `map`. Would affect relative abundance, but not beta or alpha div. May need another strategy for list vars? 
- Assuming that displayLabels are not written with meaningful delimiters (unlike entityId, varId)
- Currently we add the displayLabel if it is present at every possible location in the json output, including tables.

# Testing
All previously passing tests pass. New tests added.


d-callan (approved):
looks good. think the support for display labels will be primarily for computed vars, which im not anticipating having much if any overlap w list vars as of now (right?).. so i think its ok if list var support is pretty limited for now. we can always expand it later if needed.

asizemore:
There could be overlap based on what we do with alpha div. If our alpha-div calculation returns values per sample for all three methods, then we could be in a computed var + list var case. But I agree we can expand if needed - list var alpha div is not phase 1. 
I'll go ahead and merge!

PR #69, opened 2021-08-05
Bug/sample size table x num cat

asizemore:
_Fix for EdaDataService issue [66](https://github.com/VEuPathDB/EdaDataService/issues/66)_

# Major Updates
- `updateType` takes in both var type and shape, and if the var is a numeric categorical it converts values to characters (fixes bugs caused by numeric categorical vars).
- mosaic with `statistic=NULL` now uses `toColNameOrNull` in logic.
- logic around categorical vars that was originally based on `type='STRING'` is now based on `shape!='CONTINUOUS'`.
- `makeVariableDetails` forces variable values to be characters.
- writing `sampleSizeTable` to json now uses `toColNameOrNull` in logic (fixes funny looking `sampleSizeTable` json)

# Bug recap:
When using a numeric cat variable on the x-axis and a string cat variable on the y in the RxC viz, we get an error. The issue is that the `sampleSizeTable` does not have the expected structure:
```
{
	"sampleSizeTable": [
		{
			"size": 22567
		}
	]
}
```
but no error when using a string cat for both x and y.

# Investigation
In two places we make `sampleSizeTable` formatting decisions based on the type/shape of the x var: the first in `class-plotdata.R` and the second in `getJSON`. Originally, these decisions were made based on if the x var `dataType== 'STRING'`. However, now we have many more var types and shapes, so I believe we want the logic to be based on if the var is/is not continuous (see notes below).

During testing of the above changes, I also noticed that even though there is no error when using a string cat for both vars, the output `sampleSizeTable` json is still odd ("EUPATH_0000096.EUPATH_0010062" is the xVar):
```
"sampleSizeTable": [
    {
      "EUPATH_0000096.EUPATH_0010062": [
        "Caretaker not available after 3 visits",
        "Caretaker refused",
        "Child cannot be found",
        "Conducted",
        "Other"
      ],
      "size": [
        669,
        149,
        58,
        20981,
        683
      ],
      "xVariableDetails": {}
    }
  ],
```

This empty `xVariableDetails` arises from trying to access the xVar column of `sampleSizeTable`, but using only the varId to access the column, instead of varId.entityId. Thanks to `toColNameOrNull`, this was a simple fix, and now we have the following structure for both num cat and string vars on the x axis:
```
"sampleSizeTable": [
    {
      "xVariableDetails": {
        "variableId": "EUPATH_0000096.EUPATH_0015720",
        "entityId": "EUPATH_0000096",
        "value": [
          1,
          2,
          3
        ]
      },
      "size": [
        7094,
        8212,
        7261
      ]
    }
  ],
```

After discussion with @d-callan , we decided to handle numeric categoricals by simply converting them to strings in `updateType`. Note we do _not_ change their `dataType`, so we still can handle them uniquely if the situation arises (overlay...). In general, we force values of the numeric categorical vars to flow into plot.data as strings (see [lines 237-241](https://github.com/VEuPathDB/EdaDataService/pull/65/files)), but the `updateType` additions add an extra layer of safety.

# Testing
All previously passing plot.data tests pass except 2 in mosaic. These have to do with handling of facets/panels and naming in json and will be addressed in a new [issue](https://github.com/VEuPathDB/plot.data/issues/71). New testing implemented for handling of numcats, `sampleSizeTable` organization, and column names like 'entityId.varId'.
- All data for testing now has column names with 'entity.varId' format. 
- Reference data was updated appropriately.

# Notes
- I struggled with basing the logic on `dataShape != 'CONTINUOUS'` vs. `dataType == 'STRING' || dataShape == 'CATEGORICAL'`. Eventually I chose the former because I expect we might have ordinal or binary vars that are of type number, and we'd like them to be treated similarly. 
- Noticed while updating testing that many group functions set column names by reference. Not necessarily an issue just highlighting to ensure we still want that behavior.


asizemore:
Tested locally - original mosaic request does not err.

Screen Shot 2021-08-12 at 9 21 22 AM

d-callan:
i think once this is ready well make anoter tagged version of plot.data available

tests/testthat/test-group.R
```diff
@@ -2,120 +2,138 @@ context('group')
...
   expect_equal(nrow(dt),16)
-  expect_equal(names(dt), c('group', 'panel', 'x', 'min', 'q1', 'median', 'q3', 'max'))
+  expect_equal(names(dt), c('entity.group', 'entity.panel', 'entity.x', 'min', 'q1', 'median', 'q3', 'max'))
 })
 
 #test the more complicated ones
 test_that("groupSmoothedMean() returns an appropriately sized data.table", {
-  dt <- groupSmoothedMean(data.xy, 'x', 'y')
+
+  df <- copy(data.xy)
```
d-callan, line 49: data.table::copy

d-callan: also, why do we need to keep copying it now?

d-callan (commented):
looks good. there is the possibility if/when we get data from something other than eda that we wont have entityId.varId col headers. we should be (and i think we are) able to handle both cases. we can confirm that and test for the varId=colHeader case when we actually meet it.

id like to know why were making all these copies of the test data tables before merging

tests/testthat/test-group.R
asizemore: in groupSmoothedMean and a few other group functions, we update dt names by reference. I noticed that after changing column names in data.xy to entity.varid, that when testing `groupSmoothedMean(data.xy, 'x', 'y')` would work but then the next round of `groupSmoothedMean` got a data.xy with altered column names.

d-callan: ok. then lets make the copy in groupSmoothedMean, and then update the col names on the copy.

d-callan: or update names the old school way there and get the copy for free. either way is fine w me really. but i actively decided recently that the group fxns wont work by reference

d-callan:
unless you have something outstanding, id like to merge today. i mean to try to get everything for phase 1 on ash mon

asizemore:
nothing outstanding from me. Thank you for handling those changes! Happy to merge when you are.

PR #77, opened 2021-08-19
Improve Logging

d-callan:
first pass at timed status messages


d-callan:
worth exporting logWithTime ?

asizemore:
logWithTime is amazing and I'd vote for exporting

asizemore:
A small thing but consider adding documentation of the `verbose` parameter (noticed it was missing from bar() and bar.dt())

Suggested logging additions:
- in plot.dt() functions after arguments are validated. Something along the lines of "Creating box plot with parameters points= {points}, evilMode = {evilMode}...". This is on the top of my mind because I just had a bug that had to do with inputs :)
- Heatmap could use another log or two, for example in `validateHeatmapPD` or after `groupSplit`
- Sort of for the future, but places where we do an conversion of the data. For example if dates come in funky and we have to convert to a particular format, it could say "unaccepted date format, coercing date to DD:MM:YYYY ... UTC". I couldn't find a great place to do this in the current code, so just really an idea for the future.

Since everything above is small/a suggestion, I'll go ahead and approve. But happy to discuss more if helpful!

asizemore (approved):

d-callan:
im guessing when we get to implementing heatmaps for real theyll see a fair bit of refactoring. so im gonna hold off on logging until then too i think.

PR #78, opened 2021-08-19
Bug/73/box statsTable structure

asizemore:
_Addresses issue #73_

The box's `statsTable` originally had all the statistics wrapped into a list. Mosaic's `statsTable` instead has one column per stat (see `TwoByTwoStatsTable` in the DS). 

This PR updated `nonParametericTest` and `nonParametricTestByGroup` so that their output was one data table with statistics as columns.

## To Dos
- [ ] Add more tests

### DIFF
--- a/R/utils-stats.R
+++ b/R/utils-stats.R
@@ -330,11 +330,10 @@ nonparametricTest <- function(values, groups) {
       "parameter" = numeric(),
       "method" = character(),
       "statsError" = jsonlite::unbox(as.character(testResult[1])))
-    testResult <- list(testResult)
   } else {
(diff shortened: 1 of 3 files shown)


asizemore:
Question: When we have an overlay var, statsTable gets one row per x value and there are no columns for the overlay var. The above removes the empty overlayVariableDetails from being written to json, but should I add an option to write xVariableDetails if they are a column in the table?

d-callan:
i like this. and agree we need xVariableDetails when the stats are calculated per x var value

d-callan (approved):

PR #89, opened 2021-09-08
Fix/87/mosaic-stats

### LINKED ISSUE #87: Odds ratio and relative risk not showing up
Odds ratio and relative risk and not showing up correctly. See example below. (To recreate: GEMS1 study-> 2 x 2 plot -> X=case or control participant; Y= Giardia, by ELISA.)
Screen Shot 2021-08-26 at 4 30 17 PM

d-callan:
resolves #87 by:

1. excluding NA from the count of unique values when determining which statistic to use for mosaic plots/ validating requested statistics. i think this is fine bc the NA values are always removed from the axes vars later in the plot.data parent class anyhow, which are whats used to determine the right stats.
2. add a test for the condition.


asizemore (approved):
Looks good!

PR #91, opened 2021-09-09
Fix/90/temporarily remove validation on dataShape

### LINKED ISSUE #90: Backend: temporarily remove plot.data constraints and validation on data shape
This will let veupathdb/edadataservice#88 work as intended, and let plot requests succeed w the new constraints there. there is a related issue to properly update these in #88 here. this is a temporary measure, to see how we like the new constraints and how severe a problem categorical numbers are going to be in eda. if we dont like the new constraints we can revert this. if we do like them, then we can proceed w #88.

d-callan:
should resolve #90

### DIFF
--- a/R/class-plotdata-bar.R
+++ b/R/class-plotdata-bar.R
@@ -65,28 +65,28 @@ newBarPD <- function(.dt = data.table::data.table(),
 }
 
 validateBarPD <- function(.bar, verbose) {
-  xAxisVariable <- attr(.bar, 'xAxisVariable')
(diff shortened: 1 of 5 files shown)


d-callan:
is this a bad strategy? and if we proceed is there anything else we need to temporarily suspend?

asizemore (approved):
I think it seems reasonable until we have a clear decision on var types. The two main things I worried this might affect are support for continuous overlay and listvars. Looked to me like continuous overlay should be fine, but we could consider removing the requirement that listvar variables be continuous (line 489 of `utils.R`). Listvars aren't coming around anytime soon so there's no rush, but it'd be more for sake of consistency I suppose.

PR #92, opened 2021-09-17
Reorganize testing

### LINKED ISSUE #72: Backend: streamline plot.data testing
We're at around 1000 unit tests. Would be really helpful to make sure we're still organized in how/where/when we write tests, and to reduce the number of test datasets to maybe 2 max. For example, one test dataset would include every type of var we expect to see in eda with the expected format, and a more general dataset for testing how a non-data-service user would interact with the package.

In general, this should help us increase velocity of bug fixing because it will streamline the addition and execution of unit tests.

asizemore:
Resolves #72 

Improves organization of unit tests. 

# To dos

- [x] Simplify to one test data set
- [x] Consistently order test sections
- [x] Consistently test handling of missingness
- [x] Resolve outstanding testing errors (commented out tests in mosaic, bin)
- [x] Use `covr::report() ` to make sure we have coverage > 80% for files with currently implemented code. Files that do not meet the 80% criteria are `utils-cut.R` and `group.R`. Both contain functions not used elsewhere in the package.
- [x] Remove canaries.

# Open questions
- Should we replace numcat tests with numeric columns that have fewer than 10 values? 
- Although all values were 1 in line 32 of `test-bin.R`, they were not all equal to 1. Changed to use `all.equal` function instead of `==` based on [discussion](https://stackoverflow.com/questions/9508518/why-are-these-numbers-not-equal). Also see line 60. 

# Structure of `test.df`
Each variable type is represented at least once. Columns are named by variable type, followed by either the number of distinct values or a letter for vars where the number of unique values is not informative. Many were added for later var constraint testing.

- entity.binA, entity.binB
- entity.cat1, ..., entity.cat7
- entity.contA, ..., entity.contD
- entity.cont5, entity.cont6 (have 5 or 6, respectively, unique floats)
- entity.dateA, entity.dateB
- entity.int2, entity.int6, entity.int7, entity.int11
- entity.factor3, entity.factor6

# Proposed organization moving forward (feedback welcome!)
1. Return appropriate object with size (returns a valid plot.data barplot object)
2. check types (plot data and config of the appropriate types). 
3. valid data table (returns an appropriately sized data.table). Also includes factor checks
4. check json (eturns appropriately formatted json)
5. check missingness (returns correct information about missing data)
6. check var constraints?
10. viz-specific tests (stats)

# Future work
- Consider adding var constraints tests. For example, check that box works with int6 as x axis but not int11. 
-


tests/testthat/test-box.R
```diff
@@ -565,59 +640,62 @@ test_that("box() returns appropriately formatted json", {
...
-  expect_equal(lapply(dt_string, function(x) {length(x[[1]])}), lapply(dt_num, function(x) {length(x[[1]])}))
+#   expect_equal(nrow(dt_string), nrow(dt_num))
+#   expect_equal(names(dt_string), names(dt_num))
+#   expect_equal(lapply(dt_string, function(x) {length(x[[1]])}), lapply(dt_num, function(x) {length(x[[1]])}))
   
-})
-
+# })
 
-test_that("box.dt() returns correct information about missing data", {
```
d-callan, line 596: did this move somewhere else and i didnt notice it?

tests/testthat/test-utils.R
```diff
@@ -196,75 +220,40 @@ test_that("makePanels() does nothing if there are no facets", {
...
 
   expect_is(dt, 'data.table')
-  expect_equal(nrow(dt),data.table::uniqueN(data.binned$x))
-  expect_equal(length(dt),data.table::uniqueN(data.binned$y)+1)
+  expect_equal(nrow(dt),data.table::uniqueN(test.df$cat7))
+  expect_equal(length(dt),data.table::uniqueN(test.df$cat6)+1)
 })
 
 
-test_that("remapListVar appropriately updates map", {
```
d-callan, line 207: same question. did this move?

d-callan (commented):
in general i like it. couple thoughts:

1. should probably avoid the dot notation for anything other than methods. i know i do it a lot, and know i shouldnt. but you should probably try to avoid forming my bad habits. test.df could become testDF or test_df etc
2. we should probably consider using covr::report() or something similar, and getting coverage of all files (excluding heatmap, density or other things not yet implemented fully) above say 80% as within scope for this issue.
3. there were a couple things looked like they were removed or moved. i didnt find where they moved to, and just want to be sure we dont lose useful tests in this process.
4. if some of the canaries/ comparisons to a reference arent useful any longer (i was using them in the beginning before we had 1000+ tests) we should remove them. 
5. think the numcat thing we can use a number w fewer than 8 or 10 values. get existing tests passing that way. if it needs more work later, it can become part of the ticket to better handle numcats. in the meantime, can probably continue to ignore any references to dataShape, if/ as necessary, to get things working.

d-callan:
adding tests for var constraints is probably good

tests/testthat/test-box.R
asizemore: yup! just moved up to line 643

tests/testthat/test-utils.R
asizemore: The remapListVar function was used in one of the earlier ways we handled listvars, but is no longer used. I should also delete it from `utils.R`.

asizemore:
Thanks to covr::report(), I learned that the following functions are defined but never used. Are we happy removing these functions from the package? Some may be used in other contexts (data loading) so I don't want to delete just because plot.data doesn't use them.
- `panelOddsRatio`
- `panelRelativeRisk`
- `cut_number`
- `is.POSIXct`
- `groupSD`

The main coverage issues are in `utils-cut.R`. Other files are well covered or their sparsely covered sections are used in the heatmap.

asizemore:
Reminder to me - after PR #107 is merged, update line tests and re-request review

d-callan:
#107 is merged.

asizemore:
Line plot tests added!

d-callan (approved):

Issue #96, opened 2021-09-27
Visualization: Boxplot Outliers must always be array of array


aaronwlsong:
India ICEMR Fever Surveillance has error of "Unexpected backend type(s) encountered" in box plot. 

For example, when x= Medication provided and y= Temperature, error happens.

![image](https://user-images.githubusercontent.com/61077845/134984365-fcd184fb-0843-42f1-beb2-2b42ff5f736d.png)

aaronwlsong:
It looks good now. Closing ticket. 
![image](https://user-images.githubusercontent.com/61077845/136263338-01ab59d6-7451-442c-82da-80f721a87d14.png)

PR #98, opened 2021-10-04
Outliers need to be arrays always

asizemore:
Addresses #96 

This PR enforces that outliers are always returned as arrays. The only change needed was for the special case of 
- one x tick, no overlay or facets, and one outlier

Additionally I tested 
- one x tick but many outliers, no overlay or facet
- one x tick, many outliers, with overlay
- no overlay or facet, one row
- multiple x values but some have only 1 outlier
and all worked fine in the original code. 

## Testing
Added a new test for a single outlier on a single x tick.
Tested original variable pair with new code
Screen Shot 2021-10-04 at 10 32 17 AM
:


d-callan:
resolves #96 (hopefully that links the thing properly)
it feels a bit crazy this was necessary, but so be it

PR #102, opened 2021-10-11
Fix/101/empty response when no complete cases

### LINKED ISSUE #101: Getting 500 responses from histogram and mosaic endpoints instead of desired valid but empty response
Back end: dfalke-b
Study: India ICEMR Fever Surveillance
Subset on: Dengue, by Panbio Dengue Early ELISA = Positive

Histogram viz using main variable = Plasmodium falciparum asexual stages, by microscopy result

```
 500 Internal Server Error {"status":"server-error","message":"Unable to complete processing","requestId":"6BMaiK4IyotjWIvJ1cHszA"}
```
Works OK without the subsetting.

d-callan:
should resolve #101


R/class-plotdata-histogram.R
```diff
@@ -46,50 +46,67 @@ newHistogramPD <- function(.dt = data.table::data.table(),
...
-  veupathUtils::logWithTime('Supporting summary statistics calculated for histogram.', verbose)
-
-  if (is.null(viewport)) {
-    viewport <- findViewport(.pd[[x]], xType)
-    veupathUtils::logWithTime('Determined default viewport.', verbose)
+  if (!length(.pd[[x]])) {
+    binSlider <- list('min'=jsonlite::unbox(NA), 'max'=jsonlite::unbox(NA), 'step'=jsonlite::unbox(NA))
+    binSpec <- list('type'=jsonlite::unbox(binReportValue), 'value'=jsonlite::unbox(NA))
+    viewport <- list('xMin'=0, 'xMax'=-Inf) 
+    summary <- list('min'=jsonlite::unbox(""),
```
asizemore, line 53: Why some empty strings and some NAs? Curious if there's a general guideline or something about the expected response would dictate one over the other.

R/utils-bin.R
```diff
@@ -160,6 +167,7 @@ findBinWidth.Date <- function(x, na.rm = c(FALSE, TRUE)) {
 #' @importFrom grDevices nclass.FD
 #' @importFrom grDevices nclass.Sturges
 findNumBins <- function(x) {
+  if (!length(x)) { return(NULL) }
```
asizemore, line 170: Does this also return null (or another expected value) if all values in x are NA?

R/utils.R
```diff
@@ -223,6 +223,10 @@ pruneViewportAdjustmentFromBins <- function(bins, xVP, x, viewport) {
 }
 
 findBinStart <- function(x) {
+  if (!length(x)) {
```
asizemore, line 226: The findBinStart and end functions return what we'd expect for x all NAs as well?

tests/testthat/test-box.R
```diff
@@ -1,5 +1,45 @@
...
+  attr <- attributes(dt)
+  expect_equal(attr$completeCasesAllVars[1], 0)
+
+  map <- data.frame('id' = c('entity.binary1', 'entity.cont', 'entity.binary2'),
+                    'plotRef' = c('xAxisVariable', 'yAxisVariable', 'overlayVariable'),
+                    'dataType' = c('STRING', 'NUMBER', 'STRING'),
+                    'dataShape' = c('CATEGORICAL', 'CONTINUOUS', 'CATEGORICAL'),
+                    stringsAsFactors = FALSE)
+  df <- data.noneComplete
+
```
asizemore, line 37: may be helpful to add a test to ensure the output dt looks as expected

asizemore (approved):
A few questions/comments but nothing blocking. Approved!

R/class-plotdata-histogram.R
d-callan: i think the rule weve agreed to is for scalar numbers to return null (js)/ NA (R), for scalar strings to return an empty one and for the main data essentially an empty frame of an object.

d-callan: i was talking w bob about it yesterday

R/utils-bin.R
d-callan: trying to keep internally consistent here. binWidth and numBins within plot.data are always either a number or NULL. logic on how to provide binWidth of numBins back to the client is up to the plot class.

R/utils.R
d-callan: trouble w doing some of the work months ago is i cant quite remember this one. think im trying to make sure the data object is valid in structure but empty here. but proves your other point about tests for the data structure...

R/class-plotdata-histogram.R
asizemore: cool sounds good! Thank you!

PR #107, opened 2021-10-29
Feature/31/Line plot Y-axis aggregation functions

d-callan:
think i still need some group fxn tests probably, and to think about how i want to factor the scatter class. im thinking line plot might be its own class, w time series as a subclass.


R/class-plotdata-line.R
```diff
@@ -0,0 +1,300 @@
...
+  }
+  if (attr$xAxisVariable$dataType == 'DATE') {
+    .pd$seriesX <- lapply(.pd$seriesX, format, '%Y-%m-%d')
+  } else { 
+    .pd$seriesX <- lapply(.pd$seriesX, as.character)
+  }
+  if (attr$yAxisVariable$dataType == 'DATE') {
+    .pd$seriesY <- lapply(.pd$seriesY, format, '%Y-%m-%d')
+  } else {
+    .pd$seriesY <- lapply(.pd$seriesY, as.character)
```
asizemore, line 78: Why do we want to convert numeric to character for the y axis? Is the client converting back to numbers?

asizemore (approved):
Left one question about the y axis output. Otherwise looks good!

R/class-plotdata-line.R
d-callan: i did it bc the constraints in the data service claim to be able to handle dates currently. and r can technically get means and medians for dates. idk that to be particularly useful, but i dont know its really wrong either.

asizemore: oh i see it now. Cool

asizemore: and i see scatter is the same. Makes sense now thank you!

Issue #116, opened 2021-11-15
Backend: make sure response makes sense for collection vars


d-callan:
decided the response, when it contains a collection var, should provide var specs for the parent and child terms as well as explicitly stating it contains a collection var. need to make sure were doing this (bc i think we prev said wed try to provide the display labels rather than the specs).

asizemore:
To clarify, right now we return the following as part of the plot config:
```      "listVariableDetails": {
        "variableId": [
          "contB",
          "contA",
          "contC"
        ],
        "entityId": [
          "entity",
          "entity",
          "entity"
        ]
```
This holds information about each variable chosen as part of the collection var.

Instead, we want something like
```
"xAxisVariableDetails":
  "variableId": "myVarId"
  "entityId": "entity"

"yAxisVariableDetails":
  "variableId": "myOtherVarId"
  "entityId": "entity"
```
where `myVarId` would have the vocabulary "contA, contB, contC" and `myOtherVarId` would be the new continuous variable.

We also want to remove all display names for collection vars, and add some sort of "containsCollectionVar" flag. 

Did I understand that all correctly?

d-callan:
i think the variableId + entityId combo is a VariableSpec. so i think any axis that can represent a collection var needs to be able to return an array of VariableSpecs. 
"xAxisVariableDetails":[
  {"variableId": "myVarId"
  "entityId": "entity"},
 {"variableId": "myVarId2"
  "entityId": "entity"}]

xAxisVariableDetails i think can also probably gain some sort of flag something like 'isCollectionVariable' indicating it represents a collection var. then if it is a collection w a single child selected then its still displayed consistently w other collections. 
yAxisVariableDetails possibly needs to gain a flag indicating its the collection value. think we can infer entity id for the var spec and make something up for the var id i suppose.
this implies the data service gains a VariableDetails object w these two things as optional properties.
that leaves specifying the parent var spec somehow. that may end up being relevant to the viz generally as opposed to individual axes, im not sure. but i think maybe for that we add something like 'collectionParentVariableSpec' to the plot config.

youll need a PR here as well as the data service for this one as well, as we finalize the api.

d-callan:
think that also means the StrataVariableDetails object in the data service will likely need updating to use the new VariableDetails rather than VariableSpec

asizemore:
After discussion on 12/14, we'll assume what we have is fine and treat everything new like a bug. Closing with the expectation that we'll open many new issues on the topic soon :)

PR #117, opened 2021-11-17
Feature/115/computed var metadata api

### LINKED ISSUE #115: Backend: ensure plot config respects ComputedVariableMetadata api
check veupathdb/edadataservice#55 for the api we need to match.

asizemore:
Resolves #115 

## Details 

The current structure of "computedVariableMetadata" as defined in the data service looks like this:

```ComputedVariableMetadata:
    additionalProperties: false
    properties: 
      displayLabel:
        type: string[]
        required: false
      defaultRange:
        type: number[]
        maxItems: 2
        minItems: 2
        required: false
      vocabulary:
        type: string[]
        required: false
      isOrdinal: 
        type: boolean
        required: false
```

We're only using defaultRange and displayLabel for the near- and mid-future since planned computed vars so far are numeric, so only these properties are the two implemented and tested in this PR.

## Notes

I struggled with passing each property as an argument, or passing the whole list as an argument. I expect the computedVariableMetadata to gain more and more properties, which then combined with it being passed directly from the compute output swayed me towards just inputting the list itself. Still, not sure passing a list as a function argument is really proper so maybe it should be the other way...


asizemore:
Also is "defaultRange" the right term? Maybe should be "annotatedRange" to be more similar with var annotations?

d-callan:
ill look at this today. but i think displayRange is what the EDA* services are calling it.

i also think the list approach is the best for now. typically when you start to have those types of debates w yourself its a good sign your headed toward a new object. theres a few places in plot.data thats the case, and i think we can handle them all together at some point under some general 'refactor' issue. maybe when we switch to s4.

d-callan (commented):
pretty good. might have computed dates some day w ranges though? might need to update that here and in the api (would mean a string type rather than number). 

i also double checked the display range api for variable metadata in the subset service. its two separate properties, rather than one w two values. the two are 'displayRangeMin' and 'displayRangeMax'. i think we should match that. i also think were using 'displayName' rather than 'displayLabel'. so that i think needs updating here and in the data service api as well.

d-callan:
i guess the other thing is to make sure in the data service were actually passing this through from the compute output to plot.data. maybe open a second PR there?

asizemore:
Great. Updated language to use "displayName", "displayRangeMin", and "displayRangeMax".
Also updated output of range min and max to be strings. The compute var details will have if it's a date or number so we can use that in the front end to convert as necessary.

asizemore:
Heading to the data service to make a corresponding PR. Will link it here when done!

asizemore:
See also microbiomeComputations [PR 5](https://github.com/VEuPathDB/microbiomeComputations/pull/5) and data service [PR 134](https://github.com/VEuPathDB/EdaDataService/pull/134)

asizemore:
After discussion, we're planning to add collection var info to the computed var metadata. I'll add that to this PR. IP

asizemore:
New output example:

```
"collectionVariable": {
          "collectionType": "abundance",
          "collectionVariablePlotRef": "facetVariable2",
          "collectionValuePlotRef": "yAxisVariable",
          "collectionVariableDetails": [
            {
              "variableId": "contB",
              "entityId": "entity"
            },
            {
              "variableId": "contA",
              "entityId": "entity"
            }
          ]
        }
```

asizemore:
This PR also addresses #116 

With microbiomeComputations [PR 6](https://github.com/VEuPathDB/microbiomeComputations/pull/6), now writing collectionVariableDetails and other info in `computedVariableMetadata`. Apps will return the collectionType, then plot.data will add any collection var plot info. 

Currently keeping plot.data's collectionVar separate from computed var metadata as long as possible, since this likely a short-term solution.

See also data service [PR](https://github.com/VEuPathDB/EdaDataService/pull/134) for updates.

R/class-plotdata-beeswarm.R
```diff
@@ -27,10 +27,14 @@ newBeeswarmPD <- function(.dt = data.table::data.table(),
                          jitter = NULL,
                          median = logical(),
                          evilMode = logical(),
-                         listVarDetails = list('inferredVariable' = NULL,
+                         collectionVariableDetails = list('inferredVariable' = NULL,
                                                'inferredVarPlotRef' = NULL,
-                                               'listVarPlotRef' = NULL,
-                                               'listVarDisplayLabel' = NULL),
+                                               'collectionVariablePlotRef' = NULL,
+                                               'collectionVarDisplayLabel' = NULL),
```
d-callan, line 33: my only question is about the display label here. maybe i just missed it, but im not sure i saw it used anywhere. plus it probably makes the most sense for the client to find display labels given variable specs..

asizemore: Sounds good. collectionVarDisplayLabel removed!

PR #118, opened 2021-11-18
Fix/88/reinstate data shape

### LINKED ISSUE #88: Backend: update plot.data logic based on data shape
the data service constraints are being updated in veupathdb/edadataservice#88 as per requests in veupathdb/edadataservice#87 and veupathdb/web-eda#310 . the proposal here is to equivalently update logic in plot.data to match the new var constraints. 
i think this means wed have to:

1. pass var constraints to plot.data from the data service, so it knows the max number of allowed values as well.
2. stop passing data shape to plot.data
3. update logic and validation based on data shape to use the max allowable values instead
4. for cases where either a continuous or categorical var is allowed (ex: gradient vs categorical overlay vars) have to conditionally determine behavior based on some number of max values. the difficulty here is we dont have a good way of knowing that number, currently. the only case currently before us (color) inherently has a limit of 8 values. we could hard code that for now, but makes changing to a different categorical colormap potentially unduly difficult. that doesnt help the hypothetical case where wed like the scatter plot to behave as a beeswarm if the xaxis is inferred as categorical.

in any case, id like to wait to implement these changes, since theyre more work w potentially less gain, until we see if we like the related changes in the data service.

d-callan:
resolves #88 

decided to go w the minimalist approach to constraints at this level, given we cant well anticipate what future vizs might need. continuous overlay support is back though.

there is a bit of work remaining, which i will detail in a separate issue, that can only be done once the b54 datasets are reloaded w forceStringType implemented as an annotation rather than the automagical inference of categorical numbers.

### DIFF
--- a/R/class-plotdata-bar.R
+++ b/R/class-plotdata-bar.R
@@ -65,28 +65,6 @@ newBarPD <- function(.dt = data.table::data.table(),
 }
 
 validateBarPD <- function(.bar, verbose) {
-#  xAxisVariable <- attr(.bar, 'xAxisVariable')
-#  if (!xAxisVariable$dataShape %in% c('BINARY', 'ORDINAL', 'CATEGORICAL')) {
-#    stop('The independent axis must be binary, ordinal or categorical for barplot.')
-#  }
-#  overlayVariable <- attr(.bar, 'overlayVariable')
-#  if (!is.null(overlayVariable)) {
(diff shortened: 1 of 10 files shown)


asizemore (approved):
Cool. Stating my read of what's happening:
- dataShape constrains only axes that need better precision than just any number (ex also want dates)
- dataShape also used for non-constraint decision making. For example, gradient colormap

Also noting that our constraint on number of unique vals for facet is out of date. It's still larger than in the data service (7) and facets are in flux, so probably not really worth changing much just yet.

PR #125, opened 2022-01-11
binned x axis, error bars and categorical y axis for line plot

d-callan:
related to veupathutils/edadataservice#140

still need to incorporate the number of samples per bin. want to think about it a bit so its relatively future proof for when we start adding proportions to the mix..


asizemore:
Why does it need to also return binspec? It's part of the request, no?

R/bin.R
```diff
@@ -1,3 +1,37 @@
...
+  data$binEnd <- findBinEnd(data$binLabel)
+  data <- data[order(data$binStart),]
+  data$binStart <- as.character(data$binStart)
+
+  data <- collapseByGroup(data, group, panel)
+
+  return(data)
+}
+
+binMean <- function(data, x, y, group = NULL, panel = NULL, binWidth = NULL, viewport) {
```
asizemore, line 18: could we distinguish these two binning functions from others that take only a single variable? Maybe like binXMeanY or binMeanXY... idk something to suggest it depends on both x and y.

R/class-plotdata-line.R
```diff
@@ -52,35 +55,51 @@ newLinePD <- function(.dt = data.table::data.table(),
...
+    veupathUtils::logWithTime('Determined default viewport.', verbose)
+  } else {
+    viewport <- validateViewport(viewport, xType)
+  }
+  attr$viewport <- lapply(viewport, as.character)
+  attr$viewport <- lapply(attr$viewport, jsonlite::unbox)
+
+  # if no binWidth provided, should we find one or assume a binWidth of 0?
+  # does a binWidth of 0 really make sense, or do we need to make all of the binning stuff optional?
+  # ultimately the question is how to make binning optionaland whats the default behavior?
```
asizemore, line 76: i think we decided to have default no binning. But i guess what if the user turns on binning and sets the bin width to 0. Should that be different than just returning all the points? I suppose on one hand if no binning = binwidth 0 there's a possibility to have a more consistent response, but if they're different it's more consistent with how the viz might look from the user's perspective, so is a little more clear in that way. 

Actually if i was using plot.data outside of the eda and i just wanted to return line plot data, i would not expect no binning to be the same as binWidth 0. I'd expect to just get the series back. Not sure if that's helpful :)

R/bin.R
```diff
@@ -1,3 +1,37 @@
+binMedian <- function(data, x, y, group = NULL, panel = NULL, binWidth = NULL, viewport) {
+  data <- data[data[[x]] >= viewport$xMin & data[[x]] <= viewport$xMax,]
+  data$binLabel <- bin(data[[x]], binWidth, viewport)
+
+  byCols <- colnames(data)[colnames(data) %in% c('binLabel', group, panel)]
+  data <- data[, list(value=roundedMedian(get(..y))), by=eval(byCols)]
+
+  data$binStart <- findBinStart(data$binLabel)
+  data$binEnd <- findBinEnd(data$binLabel)
```
asizemore, line 9: just a check to make sure binLabel column gets removed somewhere

asizemore (approved):
all looks sane :)

R/utils-stats.R
```diff
@@ -1,3 +1,33 @@
+medianCI <- function(x, conf.level=.95) {
+  w <- try(wilcox.test(x, conf.int=T, conf.level=conf.level))
```
asizemore, line 2: My stats is admittedly weak. Curious why use wilcox.test to compute CIs for medians when we could use a direct [formula](https://www-users.york.ac.uk/~mb55/intro/cicent.htm)? Is it the lack of normality assumption that makes it a winner for this case?
I also ask because `wilcox.test` is also computing things we don't need, and it could be faster to compute the CIs directly.

asizemore (commented):
Are proportions still planned for this PR? Otherwise it's great!

R/bin.R
d-callan: its a good question. i was debating it w myself for a bit, particularly when i got to the proportion option for lineplot and a binProportion already existed. eventually i decided to punt on it for now. i want to think about this group of fxns more, but we need lineplot in the short term

R/class-plotdata-line.R
d-callan: yea i didnt want to have to do it but im thinking we need binning to be properly optional..

R/utils-stats.R
d-callan: .. bc it was easy to do and meant i didnt have to think too hard and i was feeling a bit lazy. a little i was also inclined to making the fxn as consistent w the other CI fxns as possible. if it turns out to be a performance problem ill revisit it, but i dont think im going to over optimize at this point.

d-callan:
think i need to add unit tests and possibly support for ordinal x-axes, as well as a warning that binWidth will be ignored for ordinal x-axes.

asizemore (commented):
Nice! A few comments/questions. Also may be worth adding a test to make sure 0/0 is handled as we expect. And adding user docs for how we calculate CIs (obviously outside of this PR just mentioning before i forget).

I was also a tad surprised to see the default denom was values not in the numerator. It's a perfectly reasonable default, i think i just expected to see all values as the default denom. Do you feel like the former will be a more common use case for the epi data?

R/bin.R
```diff
@@ -96,3 +97,34 @@ binProportion <- function(data, col, group = NULL, panel = NULL, binWidth = NULL
...
+                     denominatorCount = sum(denominatorValues %in% get(..y));
+                     list(value=roundedRatio(numeratorCount, denominatorCount),
+                          binSampleSize=proportionSampleSize(numeratorCount, denominatorCount))}, 
+                     by=eval(byCols)]
+  }
+
+  data$binStart <- findBinStart(data$binLabel)
+  data$binEnd <- findBinEnd(data$binLabel)
+  data <- data[order(data$binStart),]
+  data$binStart <- as.character(data$binStart)
```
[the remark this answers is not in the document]
d-callan: yea its do w ordering. i dont like it but findBinEnd returns character i think regardless of the actual data type while findBinStart doesnt bc we order on it later. probably an inconsistency worth sorting out at some point, but also relatively low priority in my mind and not specific to line plot

R/class-plotdata-line.R
```diff
@@ -58,9 +60,32 @@ newLinePD <- function(.dt = data.table::data.table(),
...
   group <- veupathUtils::toColNameOrNull(attr$overlayVariable)
   panel <- findPanelColName(attr$facetVariable1, attr$facetVariable2)
 
+  if (yType == 'STRING') {
+    if (is.null(numeratorValues)) {
+      stop("Numerator values must be specified for categorical y-axes.")
+    }
+    if (is.null(denominatorValues)) {
+      values <- unique(.pd[[y]])
+      denominatorValues <- values[values %ni% numeratorValues]
```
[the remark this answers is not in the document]
d-callan: no thats a typo

R/class-plotdata-line.R
```diff
@@ -72,9 +97,8 @@ newLinePD <- function(.dt = data.table::data.table(),
   attr$viewport <- lapply(viewport, as.character)
   attr$viewport <- lapply(attr$viewport, jsonlite::unbox)
 
-  # if no binWidth provided, should we find one or assume a binWidth of 0?
-  # does a binWidth of 0 really make sense, or do we need to make all of the binning stuff optional?
-  # ultimately the question is how to make binning optionaland whats the default behavior?
+  # if no binWidth is provided, find one. if the user doesnt want binning they can set binWidth to 0
+  # if someone complains about that well add a boolean param to indicate if binning is desired
```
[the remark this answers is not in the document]
d-callan: i talked to bob about it and i think at least for now if the user toggles binning off hell just pass 0 for the binWidth for numbers and 1 day for dates. thats probably not ideal, but it lets us make b56. and ill open an issue about this and the bin fxns naming and organization. the two might end up related the more i think about it, and so i dont want to touch either too close to release.

asizemore (approved):
Lovely!

PR #128, opened 2022-02-04
Feature/126/impute0s

### LINKED ISSUE #126: Mbio: add new arg to map for imputing 0s
Once imputing 0s function is ready, need to add a flag in the map so that the data service can send var-specific info to plot.data about imputing 0s.

asizemore:
Resolves #126 

This PR adds an optional 'naToZero' column to `map` which tells plot.data to replace any `NA` in that column with 0.

Depends on veupathUtils [PR #4](https://github.com/VEuPathDB/veupathUtils/pull/4)

## Notes
- Will add line plot tests after the big line plot merge
- I should also think of better tests for this in general.


R/utils-json.R
```diff
@@ -1,4 +1,4 @@
-makeVariableDetails <- function(value, variableId, entityId, displayLabel = NULL) {
+makeVariableDetails <- function(value, variableId, entityId, displayLabel = NULL, naToZero = NULL) {
```
d-callan, line 1: this is interesting.. im not sure we need to tell the client weve imputed 0s. is there something im not thinking of?

R/class-plotdata.R
```diff
@@ -49,25 +55,37 @@ newPlotdata <- function(.dt = data.table(),
...
   facet2 <- veupathUtils::toColNameOrNull(facetVariable2)
   facetType2 <- veupathUtils::toStringOrNull(as.character(facetVariable2$dataType))
   facetShape2 <- veupathUtils::toStringOrNull(as.character(facetVariable2$dataShape))
+  if (identical(veupathUtils::toStringOrNull(as.character(facetVariable2$naToZero)), 'TRUE')) {impute0cols <- c(impute0cols, facet2)}
 
+  # Replace NAs with 0s if naToZero set
+  if (!!length(impute0cols)) {
+    veupathUtils::setNaToZero(.dt, cols=impute0cols)
+    veupathUtils::logWithTime(paste('Replaced NA with 0 in the following columns: ', impute0cols), verbose)
+  }
```
d-callan, line 88: we should validate these.. check before actually attempting the imputation that we havent done things like request imputed 0s on a string or anything similar.

tests/testthat/test-box.R
```diff
@@ -712,6 +712,26 @@ test_that("box.dt() returns correct information about missing data", {
...
   dt <- box.dt(df, map, points = 'none', mean = FALSE, computeStats = TRUE, evilMode = TRUE)
   expect_equal(attr(dt, 'completeCasesAxesVars')[1], sum(!is.na(df$entity.contB) & !is.na(df$entity.cat4))) 
+
+
+  ## Using naToZero to change some NAs to 0
+  map <- data.frame('id' = c('entity.cat3', 'entity.contB', 'entity.cat4'),
+                    'plotRef' = c('overlayVariable', 'yAxisVariable', 'xAxisVariable'),
+                    'dataType' = c('STRING', 'NUMBER', 'STRING'),
+                    'dataShape' = c('CATEGORICAL', 'CONTINUOUS', 'CATEGORICAL'),
+                    'naToZero' = c('FALSE', 'TRUE', ''), stringsAsFactors=FALSE)
```
d-callan, line 722: should update any relevant documentation (particularly our readme) w what the acceptable values are here, what happens if we pass the empty string or TRUE for a string type etc.

R/utils.R
```diff
@@ -49,6 +50,7 @@ plotRefMapToList <- function(map, plotRef) {
   dataType <- veupathUtils::toStringOrNull(map$dataType[map$plotRef == plotRef])
   dataShape <- veupathUtils::toStringOrNull(map$dataShape[map$plotRef == plotRef])
   displayLabel <- veupathUtils::toStringOrNull(map$displayLabel[map$plotRef == plotRef])
+  naToZero <- veupathUtils::toStringOrNull(map$naToZero[map$plotRef == plotRef])
```
d-callan, line 53: actually this might be the place to do the validating?

R/class-plotdata.R
```diff
@@ -49,25 +55,37 @@ newPlotdata <- function(.dt = data.table(),
                          ...,
                          class = character()) {
 
+  impute0cols <- c()
```
d-callan, line 58: not essential, but i feel like this makes me want a generic helper that can take a vector of our per-variable plot reference list things (another class we could use when we eventually refactor) and some predicate, and return the column names that are associated w the variables meeting the condition..

`impute0cols <- findColNamesByPredicate(c(xAxisVariable,yAxisVariable,...), function(x){x$naToZero == 'TRUE' })`

tests/testthat/test-box.R
d-callan: also might need to test for an NA in addition to empty strings

R/class-plotdata.R
d-callan: might mean first writing/ finding a helper to identify which members of some vector meet the condition, similar to purrr::every except instead of a single T/F you get a vector of them. then wrap this more specific fxn around that more general one..

d-callan: or just do this maybe `unlist(lapply(list(xAxisVariable,yAxisVariable,zAxisVariable), function(x){if(x$naToZero == T) x$id}))`

up to you.

R/utils-json.R
asizemore: the main place i was thinking of is maybe in the complete cases table, or some other annotation on the plot somewhere that says we imputed zeros. For example, in reality variable A has 50 missing values, but now it looks like it has none. That could be confusing if there wasn't a little asterisk or something to indicate that we've imputed 0s for this variable.

d-callan: sure, but even if we go that route, i think the client probably already knows this? i mean its a decision made per variable based on variable metadata, which the client has access to.

R/utils.R
```diff
@@ -68,6 +69,18 @@ plotRefMapToList <- function(map, plotRef) {
...
+  
+  # Validate naToZero and fix if necessary
+  if (length(naToZero) == 0) {
+    naToZero <- ''
```
d-callan, line 76: so looks like all we test for is a TRUE value.. does that mean that NA and empty strings are functionally treated as FALSE? i think thats a reasonable default (though we should probably provide a warning here when we see these input values from the user), but i also think we only need to store them internally as FALSE. im not sure i understand the need to keep an empty string as part of our internal structure here when we could simply require naToZero to be a logical at this level.. is there something special about the empty string i didnt notice?

d-callan (commented):
very close now

R/utils.R
```diff
@@ -68,6 +69,20 @@ plotRefMapToList <- function(map, plotRef) {
...
+    warning("Encountered empty naToZero value. Setting naToZero = FALSE.")
+    naToZero <- FALSE
+  } else if (is.na(naToZero) || naToZero=='') {
+    warning("Encountered '' or NA as the naToZero value. Setting naToZero = FALSE.")
+    naToZero <- FALSE
+  } else {
+    if (identical(naToZero, 'TRUE')) naToZero <- TRUE
+    if (identical(naToZero, 'FALSE')) naToZero <- FALSE
+  }
+  if (!identical(naToZero, '') & !(naToZero %in% c(TRUE, FALSE))) {stop("plotRefMapToList error: Unrecognized value submitted for naToZero")}
```
d-callan, line 85: at this point shouldnt naToZero always be T/F ? whats the check for empty string here doing?

README.md
```diff
@@ -23,6 +23,8 @@ All `plot.data` functions require at least the following arguments:
     - `plotRef` The plot element to which that variable will be mapped. Options are `xAxisVariable`, `yAxisVariable`, `zAxisVariable`, `overlayVariable`, `facetVariable1`, `facetVariable2`. 
     - `dataType` Options are `NUMBER`, `INTEGER`, `STRING`, or `DATE`. Optional.
     - `dataShape` Options are `CONTINUOUS`, `CATEGORICAL`, `ORDINAL`, `BINARY`. Optional.
+    - `naToZero` Options are `TRUE`, `FALSE`, or `''`. Optional. Indicates if NAs found in that variable should be replaced with 0s. Only numeric columns can have `naToZero=TRUE`.
```
d-callan, line 26: should maybe explain empty string, NA and NULL are assumed to mean FALSE

d-callan (approved):
just a couple little things that would make clearer to a user what were doing

PR #137, opened 2022-03-10
Confidence intervals bug

### LINKED ISSUE #136: Backend: Median 95% CIs should represent real values from the data rather than estimated hypothetical ones
bob noticed a thing:

> I checked the actual raw response values and for example the 15th bin has a seriesY value of 77.2 but the error bars are both greater than this
> {
> 		"lowerBound": 77.8,
> 		"upperBound": 79,
> 		"error": ""
> }

and it turns out i need to do math..

d-callan:
resolves #136


R/utils-stats.R
```diff
@@ -1,31 +1,49 @@
 medianCI <- function(x, conf.level=.95) {
-  w <- try(wilcox.test(x, conf.int=T, conf.level=conf.level))
-  if (veupathUtils::is.error(w)) {
-    ci <- list('lowerBound'=jsonlite::unbox(NA), 'upperBound'=jsonlite::unbox(NA), 'error'=jsonlite::unbox(w[1]))
+  x <- x[order(x)]
+  n <- length(x)
+  z <- qnorm(p=(1-conf.level)/2, lower.tail=FALSE)
+  j <- ceiling(.5*n - z*sqrt(n*.25))
+  k <- ceiling(.5*n + z*sqrt(n*.25))
```
asizemore, line 6: A note for later - this math would be excellent in the documentation :)

R/utils-stats.R
```diff
@@ -1,31 +1,49 @@
...
-  b <- try(binom.test(numerator, denominator, conf.level=conf.level))
-  if (veupathUtils::is.error(b)) {
-    ci <- list('lowerBound'=jsonlite::unbox(NA), 'upperBound'=jsonlite::unbox(NA), 'error'=jsonlite::unbox(b[1]))
+  p <- numerator/denominator
+  n <- denominator
+  z <- qnorm(p=(1-conf.level)/2, lower.tail=FALSE)
+  lowerBound <- p - z*sqrt(p*(1-p)/n)
+  lowerBound <- ifelse(lowerBound < 0, 0, lowerBound)
+  upperBound <- p + z*sqrt(p*(1-p)/n)
+  upperBound <- ifelse(upperBound > 1, 1, upperBound)
```
asizemore, line 41: standard NA check - if upperBound/lowerBound is NA, will `ifelse` return NA here? I was reading about it but it wasn't totally clear.

asizemore: looks like your tests below cover this case. Nvm

R/utils-stats.R
```diff
@@ -1,31 +1,49 @@
...
 }
 
 proportionCI <- function(numerator, denominator, conf.level=.95) {
-  b <- try(binom.test(numerator, denominator, conf.level=conf.level))
-  if (veupathUtils::is.error(b)) {
-    ci <- list('lowerBound'=jsonlite::unbox(NA), 'upperBound'=jsonlite::unbox(NA), 'error'=jsonlite::unbox(b[1]))
+  p <- numerator/denominator
+  n <- denominator
+  z <- qnorm(p=(1-conf.level)/2, lower.tail=FALSE)
+  lowerBound <- p - z*sqrt(p*(1-p)/n)
```
asizemore, line 38: more math for the docs :)

asizemore (approved):
Just notes and questions. Looks great! Yay math!

R/utils-stats.R
d-callan: something like NA < 0 should return NA, and something like numeric() < 0 should return logical(). both should be caught by the if condition immediately below.

PR #141, opened 2022-03-11
Update arguments to showMissingness

### LINKED ISSUE #139: Backend: make showMissingness options 'all/strata/noVars' instead of 'T/F'
support for vizs embedded in map markers

d-callan:
resolves #139 

logic is there and existing tests passing, documentation updated. think i still need to add some tests for the 'allVariables' option.


R/class-plotdata.R
```diff
@@ -192,13 +195,21 @@ newPlotdata <- function(.dt = data.table(),
   completeCasesAxesVars <- jsonlite::unbox(nrow(.dt[complete.cases(.dt[, c(x,y), with=FALSE]),]))
   veupathUtils::logWithTime('Determined total number of complete cases across axes and strata vars.', verbose)
 
-  if (evilMode) {
+  if (isEvil) {
     # Assign NA strata values to 'No data', with the exception of continuous overlays which should stay numeric
+    # TODO at some point i want to make sure our logic here is solid. 
+    # is it possible for a continuous overlay to represent something other than color for ex? and what then?
```
asizemore, line 201: Maybe it will depend on what we put in the 'overlay' vs 'z variable' bins. For example, would we count size of points in a bubble plot a second overlay or z axis? The correllegram is basically the same as well.

tests/testthat/test-pie.R
```diff
@@ -279,6 +279,6 @@ test_that("pie.dt() returns correct information about missing data", {
   # number of completeCases should be <= complete cases for each var
   expect_equal(all(attr(dt, 'completeCasesAllVars')[1] <= completecasestable$completeCases), TRUE)
   expect_equal(attr(dt, 'completeCasesAxesVars')[1] >= attr(dt, 'completeCasesAllVars')[1], TRUE)
-  dt <- pie.dt(df, map, value='count', evilMode = TRUE)
+  dt <- pie.dt(df, map, value='count', evilMode = 'strataVariables')
```
asizemore, line 282: For pie do we need to test 'allVariables' since the map will have evilMode for the x axis, assuming that's still the plan?

asizemore (commented):
exceptional.

asizemore (approved):
Lovely!

Issue #144, opened 2022-03-26
lineplot extreme edge case returns non-array for data.seriesY


bobular:
When a very large bin width is provided to the back end, and a single bin's worth of data is returned, seriesY is a scalar, instead of an array.

I realise that in general usage, the back end recommends binSlider values that give at least two bins, but there are scenarios where the user changes from months to years and the year-based constraints are unknown at the time the request is made.

e.g. GEMS1 - Birth date, bin on months and max the slider to 49. Then switch to year. 

```
{
	"data": [
		{
			"seriesX": [
				"2003-01-01 - 2052-01-01"
			],
			"seriesY": "2.8457",
			"binSampleSize": [
				{
					"N": 9436
				}
			],
			"binStart": [
				"2003-01-01T00:00:00"
			],
			"binEnd": [
				"2052-01-01T00:00:00"
			]
		}
	]
}
```

asizemore:
Addressed by #149 . Closing.

nkittur-uga:
Tried this in GEMS1 and I am unfortunately still getting the big pink error. 
![Screen Shot 2022-04-12 at 2 01 55 PM (2)](https://user-images.githubusercontent.com/67710160/163025401-2aa9e0fd-23f8-451c-9d7a-dd789d6d3925.png)

d-callan:
There was an issue w deploying some backend services which was very recently (last 5-10 min) resolved. Could you try again please?

nkittur-uga:
No errors now. Passes QA.

PR #149, opened 2022-04-06
wrap single data point in single group in list for scatter to meet api

### LINKED ISSUE #135: Scatterplot response with one point is not an array
Noticed when working on the alpha div app with GEMS

Screen Shot 2022-03-04 at 12 28 28 PM

The x-axis variable here must only have one value. So the `seriesX` and `seriesY` are returning a single number, not an array with a single number.

asizemore:
Resolves #135 

This PR updates `getJSON` so that it wraps scatterplot single data points of single groups in a list. This ensures that when the data is written to json in the eda it matches the api (array of strings, not a string).

Also considered:
- Does this issue affect the line plot as well? No, the line plot errs if given a single point. I'll make a separate issue.
- Is the json normal if some but not all scatterplot dt rows have a single point? Yes
- Is the json normal if there are multiple rows but all have only one data point? Yes

Testing - looks good!

Screen Shot 2022-04-06 at 4 36 54 PM

### DIFF
--- a/R/utils-json.R
+++ b/R/utils-json.R
@@ -165,12 +165,12 @@ getJSON <- function(.pd, evilMode) {
     namedAttrList$computedVariableMetadata <- computedVariableMetadata
   }
   
-  # Fix single data points in a single group for scatterplot - should have seriesX AND seriesY columns
-  if (nrow(.pd) == 1 && 'seriesX' %in% names(.pd)) {
-    # Fix only needed when we hae excatly one data point in our one group
-    if (length(.pd$seriesX) == 1) {
-      .pd[, c('seriesX','seriesY') := list(seriesX = list(seriesX), seriesY = list(seriesY))]
-    }
+  # Fix single data points in a single group for scatterplot and line plot - should have seriesX AND seriesY columns
+  # Fix only needed when we hae excatly one data point in our one group
+  if (nrow(.pd) == 1 && 'seriesX' %in% names(.pd) && length(.pd$seriesX) == 1) {
+    # Either one (see issue 144) or both columns may be single strings not lists. 
+    if (typeof(.pd$seriesX) != 'list') .pd$seriesX <- list(list(.pd$seriesX))
+    if (typeof(.pd$seriesY) != 'list') .pd$seriesY <- list(list(.pd$seriesY))
   }
   
   outList <- list(class = list('data'=.pd, 'config'=namedAttrList))
--- a/R/utils-json.R
+++ b/R/utils-json.R
@@ -165,7 +165,16 @@ getJSON <- function(.pd, evilMode) {
     namedAttrList$computedVariableMetadata <- computedVariableMetadata
   }
   
+  # Fix single data points in a single group for scatterplot - should have seriesX AND seriesY columns
+  if (nrow(.pd) == 1 && 'seriesX' %in% names(.pd)) {
+    # Fix only needed when we hae excatly one data point in our one group
+    if (length(.pd$seriesX) == 1) {
+      .pd[, c('seriesX','seriesY') := list(seriesX = list(seriesX), seriesY = list(seriesY))]
+    }
+  }
+  
   outList <- list(class = list('data'=.pd, 'config'=namedAttrList))
+
   if (!inherits(sampleSizeTable, 'function')) {
     outList$sampleSizeTable <- sampleSizeTable
   }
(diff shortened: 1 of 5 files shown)


asizemore:
ahh perhaps i could solve #144 here also

asizemore:
updated so that we check seriesX and seriesY individually. Bob's example in #144 had only one column not an array.

I tried to recreate his steps with clinepi backend and looking good! This is GEMS1 Case Control

Screen Shot 2022-04-06 at 6 12 53 PM

asizemore:
@d-callan ready for your review! Addresses #135 and #144 

Also noting (to me) that Rserve should not be updated until evilMode options have been changed in the data service.

asizemore:
Also i need to add a test for line plot.

R/utils-json.R
```diff
@@ -165,7 +165,16 @@ getJSON <- function(.pd, evilMode) {
     namedAttrList$computedVariableMetadata <- computedVariableMetadata
   }
   
+  # Fix single data points in a single group for scatterplot and line plot - should have seriesX AND seriesY columns
+  # Fix only needed when we hae excatly one data point in our one group
+  if (nrow(.pd) == 1 && 'seriesX' %in% names(.pd) && length(.pd$seriesX) == 1) {
+    # Either one (see issue 144) or both columns may be single strings not lists. 
+    if (typeof(.pd$seriesX) != 'list') .pd$seriesX <- list(list(.pd$seriesX))
+    if (typeof(.pd$seriesY) != 'list') .pd$seriesY <- list(list(.pd$seriesY))
+  }
```
d-callan, line 174: are we VERY confident this only impacts scatter and line? not only is this a highly specific solution, but it references seriesX and seriesY directly outside of the classes to which they conceptually belong.

tests/testthat/test-scattergl.R
```diff
@@ -677,6 +677,23 @@ test_that("scattergl() returns appropriately formatted json", {
...
+                    'dataShape' = c('CONTINUOUS', 'CONTINUOUS', 'CATEGORICAL', 'CONTINUOUS'), 'displayLabel' = c('yDisplay', 'xDisplay', 'panelDisplay', 'zDisplay'), stringsAsFactors=FALSE)
+  
+  df <- as.data.frame(testDF)
+  df <- df[1, ]
+
+  dt <- scattergl.dt(df, map, 'raw')
+  outJson <- getJSON(dt, FALSE)
+  jsonList <- jsonlite::fromJSON(outJson)
+  expect_equal(typeof(jsonList$scatterplot$data$seriesX), 'list')
+  expect_equal(typeof(jsonList$scatterplot$data$seriesY), 'list')
```
d-callan, line 696: id write tests for more than just line and scatter i think

R/utils-json.R
asizemore: yeah this solution was sort of my best of all the bad solutions i could think of. From the R perspective, the data in the .pd object looks the same whether it's this specific case of a single data point or a single data point in one of many groups. To me this meant I shouldn't put the fix within the plot class, because then i'd get out data that looks different in these different cases (one where we have a string wrapped in a list and one just a string). Which left making a fix in getJSON that annoyingly has to know about these specific columns.

Perhaps a more generalized solution would be that getJSON is passed info about columns and the api it should match? So then it would just check to see if the columns exist and that they are lists/characters/numbers/etc if they should be.

None of the above was an answer to your question :) which i'll go look into

asizemore: Finally answer to your question:
Tested all with one point and did not see the same issue in any other plot. Mosaic actually erred because of the chi-squared test. Otherwise all gave good responses.
Tested histo also with many data points but giant bin width to replicate how Bob arrived at his line plot error. All looked good.

This actually makes things odder... i did not expect those results. I'll keep thinking on this

I also thought about ensuring every non-strata non-error column in the data part of the output is a list if it's just one value, but r2 needs to be a string not an array.

d-callan: three thoughts:

1. do we know what it is making scatter and line / seriesX/Y special for this case?

asizemore: Oof finally found it! Explains both scatter and line. 
`lapply(.pd$seriesY, as.character)`
only happens for seriesY in line, and happens for both seriesX and Y in scatter. 
I'll fix it here!

asizemore:
all fixed up! @d-callan back to you!

d-callan (approved):

PR #151, opened 2022-04-12
geolocation viewport for pie

### LINKED ISSUE #150: Backend: respect geolocation viewport for pieplot
related to veupathdb/edadataservice#166 . plot.data needs to be updated to reflect the new api and to properly determine completeCases* counts

d-callan:
resolves #150


R/class-plotdata.R
```diff
@@ -86,6 +101,7 @@ newPlotdata <- function(.dt = data.table(),
     veupathUtils::logWithTime(paste('Replaced NA with 0 in the following columns: ', impute0cols), verbose)
   }
 
+  #lat and lon must be used w a geohash, so they dont need to be part of completeCases*
```
asizemore, line 104: i dont follow this part - why does being used with a geohash make a difference here?

R/utils.R
```diff
@@ -229,6 +242,30 @@ adjustToViewport <- function(x, viewport) {
...
 
+#' Filter Data Table to Geolocation Viewport
+#' 
+#'
+#' This function will filter a data.table to the specified 
+#' latitude and longitude values.
+#' @param .dt data.table with columns for latitude and longitude
+#' @param latitude Character vector specifying name of latitude column
+#' @param longitude Character vector specifying name of longitude column
+#' @param viewport List of lists specifying the geolocation viewport
```
asizemore, line 253: also may be helpful to specify list names that are expected, like xMin and left/right.

R/utils.R
```diff
@@ -229,6 +242,30 @@ adjustToViewport <- function(x, viewport) {
...
+#' @return data.table of filtered values 
+#' @export
+filterToGeolocationViewport <- function(.dt, latitude, longitude, viewport) {
+  if (is.null(viewport)) { return(.dt) }  
+
+  .dt <- .dt[.dt[[latitude]] >= viewport$latitude$xMin & .dt[[latitude]] <= viewport$latitude$xMax]
+  if (viewport$longitude$left < viewport$longitude$right) {
```
asizemore, line 260: Note: here we keep data on the viewport boundaries (with <= and => ) but in the [data service](https://github.com/VEuPathDB/EdaDataService/blob/f1883ce9e71c5bbdaedeb5e9e42007424fa298cf/src/main/java/org/veupathdb/service/eda/ds/plugin/pass/MapPlugin.java#L110-L115) we do not

asizemore (commented):
After reading this PR i'm also in favor of making map markers their own subclass :)

R/class-plotdata.R
d-callan: my understanding from bob is that if there is a geohash, there is latitude and longitude.

R/utils.R
d-callan: the logic is reversed bc one is inclusive and the other exclusive. result should be the same.

R/class-plotdata.R
d-callan: tbh this assumption in combination w the fact that any time we use geohash we seem to also need lat and lon make me think the three might be a candidate for a class when we switch to s4.

asizemore (approved):
looks good to me!

PR #155, opened 2022-04-14
handle low cardinality continuous overlay vars like categorical

### LINKED ISSUE #134: overlayVariableDetails malformed in some back end responses for low-cardinality numeric variables
Related to VEuPathDB/web-eda#914 

Not for b56 at this late hour!

Here is a GEMS1 barplot for country, overlay = categorical wealth

image

Looks like the overlayVariableDetails object is missing in the response:

```
{
	"0": {
		"PCO_0000024.EUPATH_0000143": 2,
		"label": [
			"Bangladesh",
			"India",
			"Kenya",
			"Mali",
			"Mozambique",
			"Pakistan",
			"The Gambia"
		],
		"value": [
			759,
			795,
			826,
			820,
			385,
			722,
			520
		]
	}
}
```

asizemore:
Resolves #134 

The solution implemented here is a combination of checking number of unique values as well as adding an attribute to the plot class. I went this route because it's most natural to find the number of unique values using the original input data table (before collapsing by group), and more funky to find the number of unique vars by the time we're in the writing json stage.

### DIFF
--- a/R/class-plotdata-scatter.R
+++ b/R/class-plotdata-scatter.R
@@ -58,7 +58,7 @@ newScatterPD <- function(.dt = data.table::data.table(),
   group <- veupathUtils::toColNameOrNull(attr$overlayVariable)
   panel <- findPanelColName(attr$facetVariable1, attr$facetVariable2)
 
-  if (identical(attr$overlayVariable$dataShape,'CONTINUOUS') && useGradientColorscale) {
+  if (useGradientColorscale) {
     .pd$overlayMissingData <- is.na(.pd[[group]])
     series <- collapseByGroup(.pd, group = 'overlayMissingData', panel)
     .pd$overlayMissingData <- NULL
--- a/R/class-plotdata-scatter.R
+++ b/R/class-plotdata-scatter.R
@@ -84,7 +84,7 @@ newScatterPD <- function(.dt = data.table::data.table(),
   if (class(series$seriesY) != 'list') series$seriesY <- list(list(series$seriesY))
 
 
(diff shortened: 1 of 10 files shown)


R/class-plotdata-scatter.R
```diff
@@ -81,7 +83,8 @@ newScatterPD <- function(.dt = data.table::data.table(),
   }
   if (class(series$seriesY) != 'list') series$seriesY <- list(list(series$seriesY))
 
-  if (identical(attr$overlayVariable$dataShape,'CONTINUOUS')) {
+
+  if (identical(attr$overlayVariable$dataShape,'CONTINUOUS') && useGradientColorscale) {
```
d-callan, line 87: This seems like maybe it could just test for useGradientColorscale since that condition already tested for continuous vars?

d-callan (approved):
Generally fine w this as a short term solution. Think the new issue we open should be mentioned here or in the b57 issue though. Also a thought I had was that eventually the unique values count will come from the data service and presumably be handled by plotRefMapToList. So if we wanted to set ourselves up for the future we could consider adding the unique values count to the variable details lists, but not essential.

asizemore:
Sounds good thank you! If i have time today i'll add to the map! Otherwise i'll just merge to get the fix in.

The issue for the real fix is #156

R/class-plotdata-scatter.R
asizemore: Yes! And brilliant this highlighted i had missed accounting for the rare case when there is one continuous overlay supplied as a collection variable. Many thanks to you and to our unit tests for catching this!

PR #159, opened 2022-04-18
force box labels to be strings

### LINKED ISSUE #158: Boxplot: low cardinality numeric vars have numeric box labels
Screen Shot 2022-04-18 at 6 18 50 PM

Think 'label' is always supposed to be a string, based on the api.

asizemore:
Resolves #158 

I fear we may have more fallout yet from allowing numeric vars in these classically categorical places...


d-callan:
We're confident this only affects box? I don't remember what bar is doing for ex.

asizemore:
Bar has a line to ensure its x axis is a string, so it's good. I added a test just to help sleep at night.

Quickly checked the rest of the vizs to make sure they weren't erring with a numeric var - all looking good! 

Also to answer your question from slack, it's not an api change, the box response just isn't meeting the ds api anymore because we're now getting numeric vars sometimes and they're still numbers when they get to us instead of strings.

d-callan (approved):

PR #166, opened 2022-05-21
Continuous map markers

d-callan:


d-callan:
numBinsToBinWidth is maybe 99% there, w one unit test that doesnt pass yet and im struggling w. but at least when its off its off in the right direction.. id rather pass one fewer bin than one additional one since well be coloring by them in the map.

otherwise im just hoping this is coherent... my focus recently isnt what it usually is

asizemore:
@d-callan and @bobular still getting back into work let me make sure i'm clear on the plan - so for the map markers we're always returning the same data regardless of if the var is categorical or continuous. The front end will decide to use a pie vs a bar based on the var type.

d-callan:
Think pie vs bar is a user choice. The client just needs to decide which colormap to use based on the var shape. I think of it's continuous or ordinal we use a sequential colormap and if it's a simple categorical then we use the qualitative colormap.

d-callan:
And the sequential map needs discretized based on the number of resulting bins, so maybe rankedValues is useful for continuous vars as well and should just be a requirement for the map markers overlay response.

R/utils.R
```diff
@@ -87,10 +87,10 @@ plotRefMapToList <- function(map, plotRef) {
   # Validate naToZero and fix if necessary
   # NOTE failing to set naToZero will result in a default value of FALSE
   if (length(naToZero) == 0) {
-    warning("Encountered empty or NULL naToZero value. Setting naToZero = FALSE.")
+    #warning("Encountered empty or NULL naToZero value. Setting naToZero = FALSE.")
```
asizemore, line 90: are these intentionally commented out for moving forward or just helpful for testing?

tests/testthat/test-utils.R
```diff
@@ -1,5 +1,63 @@
...
+  
+  binWidth <- numBinsToBinWidth(testDF$entity.contA, 8)
+  viewport <- findViewport(testDF$entity.contA, 'NUMBER')
+  x <- bin(testDF$entity.contA, binWidth, viewport)
+  expect_equal(data.table::uniqueN(x),8)
+
+  binWidth <- numBinsToBinWidth(testDF$entity.contA, 27)
+  viewport <- findViewport(testDF$entity.contA, 'NUMBER')
+  x <- bin(testDF$entity.contA, binWidth, viewport)
+  #expect_equal(data.table::uniqueN(x),27)
```
asizemore, line 22: fiddled around a bit to try to find a pattern. Only learned that swapping contA for contB is even weirder

asizemore (commented):
a few questions but overall looks good to me!

bobular (approved):
Got my head around most of class-plotdata-map-markers.R and think it looks OK.

R/class-plotdata-map-markers.R
```diff
@@ -38,100 +33,143 @@ newPiePD <- function(.dt = data.table::data.table(),
...
+      rankedValues <- ranked[[x]]
+      attr$rankedValues <- rankedValues
+    }
   }
-  attr$rankedValues <- rankedValues
 
-  if (is.null(viewport)) {
-    viewport <- findGeolocationViewport(.pd, lat, lon)
+  if (is.null(geolocationViewport)) {
+    geolocationViewport <- findGeolocationViewport(.pd, lat, lon)
```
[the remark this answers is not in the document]
d-callan: think for other viz types we always return the viewport we used to find stats and aggregate values etc in the response explicitly, even if it essentially just results in us saying that we used all the data.. i was trying to keep mapMarkers consistent w the others in that regard.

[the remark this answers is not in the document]
d-callan: (think youre right about finding the viewport though. ill think on it.)

R/class-plotdata-map-markers.R
```diff
@@ -91,26 +89,35 @@ newMapMarkersPD <- function(.dt = data.table::data.table(),
       xRange <- findViewport(.pd[[x]], xType)
       .pd[[x]] <- as.character(bin(.pd[[x]], binWidth, xRange))
       veupathUtils::logWithTime('Successfully binned continuous x-axis.', verbose)
+
+      # maybe worth at some point a fxn getRankedValues(x = character(), maxNumValues = integer(), otherBin = c(T,F))
+      # if maxNumValues was exceeded and otherBin = F then it would just put out an error
+      # maxNumValues could be NULL by default maybe?
+      # would something like that be better here or veupathUtils (could mbio use it?)
```
asizemore, line 96: Oh i like it. Not sure mbio has a use for it at the moment, but sounds more like a general use function, especially with defaults maxNumValues=Null and otherBin=F.

asizemore (approved):
yay maps!

PR #167, opened 2022-05-24
Container for running tests

bobular:
It might be cleaner to not install rserve,

And create the working directory as `/home/plot.data` or similar?

### DIFF
--- a/.dev/runTests.R
+++ b/.dev/runTests.R
@@ -1,0 +1,19 @@
+#
+# you can run this in a R container...
+# see ./test-container/README
+#
+
+# Handle plot.data environment (time zone, for example)
+library(dotenv)
+load_dot_env(file=".dev/.env")
+
+# Helpful for development
+library(devtools)
+library(sloop)
+
+# Load local library. If for ex we're already in the plot.data dir, will load the local plot.data
+devtools::load_all()
+
+# test
+devtools::test()
+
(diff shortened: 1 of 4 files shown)


bobular:
> i think the general idea is a strong one. a couple things worth pointing out though:
> 
>     1. the tests are run as a GitHub action on commits to any branch so theyre not to my mind high priority to include in .dev
 
Sure, but running the tests was just the first step to be able to run bits of code interactively in the correct/appropriate R environment while doing PR reviews.  

>     2. if youre doing dev you dont really want to install the package in the docker container.

Right, yes. I've been playing with this. I can modify code in my plot.data directory (which is mounted into the container), but have to run `devtools::load_all()` at my container's R prompt before it will take effect.  How do you get R to use a package "live" from a directory?

>     3. presumably all of our r packages would benefit by something similar, so i was wondering if veupathUtils isnt a better place for this dockerfile (something i had mentioned to @asizemore but i dont mind who actually does the work if youd rather)

Yes, this seems like a good idea - Ann or I can take care of that. I'd better actually do the map markers PR review rather than get bogged down in all this though!

asizemore:
@bobular this is brilliant! I should have some time next week i think so i can get started moving this to veupathUtils. 

> How do you get R to use a package "live" from a directory

As far as i know there's no way to do this, unfortunately :(

bobular:
> @bobular this is brilliant! I should have some time next week i think so i can get started moving this to veupathUtils.
> 
> > How do you get R to use a package "live" from a directory
> 
> As far as i know there's no way to do this, unfortunately :(

Great. Thanks.

The binding of the `plot.data` repo directory into the container in the `docker run` command in the README is a bit opaque. I should have commented better.

Issue #163, opened 2022-04-27
Backend: discrete colormap support in all plot types


d-callan:
based on conversation in slack it seems we might have a need (to be confirmed/ finalized yet but i wanted to write it down while i was thinking of it) in the mapveu application for all vizs to be able to discretize continuous variables for the overlay. in eda generally though continuous variables are going to use a proper gradient, possibly excepting low cardinality continuous variables. I think that logic should possibly be clarified in a dataviz meeting before implementing this. but in either case i think continuous variables w more than some number of values (8, 10, ?) need to be able to support optionally either gradient or sequential discrete colormaps, and in the second case we need to bin the data into some user defined number of equidistant ranges, w the default yet tbd. hopefully any of that made sense..

bobular:
I think all the summary stats (everything except scatterplot?) would still need to discretise if overlay stratifying a continuous variable.  
For example imagine if this histogram was stratified on Age (here simulated using "age group" already in 3 bins)
![image](https://user-images.githubusercontent.com/308639/165620721-b879e125-17e6-4366-a2ba-56ff5aea1df2.png)

bobular:
If a scatterplot is "overlaid" with a continuous variable, this is only possible without trend lines, right?

d-callan:
Think that's what we did but @asizemore  could confirm. The other option would be to provide a single trend line per panel. I don't have strong opinions.

asizemore:
> If a scatterplot is "overlaid" with a continuous variable, this is only possible without trend lines, right?

@bobular correct! The scatterplot should return a nice error saying exactly that if you try to give it a continuous overlay and request trend lines.

Sorry i'm behind on all the other discussions on this topic! I'll try to get up to speed and provide better thoughts tomorrow!

asizemore:
I'm proud to say i am up-to-date on the slack thread and mockups! This is going to be so cool!

Right, so if scatterplot has a continuous overlay that is not binned, so it's using the continuous gradient colormap, then trend lines don't really make sense. If the continuous overlay gets binned, so using the discretized gradient colormap, then i'm not terribly opposed to trend lines. Just a little opposed 🙂 . I'm not crazy about doing math on the binned data and then deriving meaning from the differences in those trend lines, i guess. But i dont have any math to really back that up, just a general queasiness.

bobular:
> But i dont have any math to really back that up, just a general queasiness.

Yeah I agree. Can't put my finger on it, but it seems sketchy to me too.

[...]

PR #171, opened 2022-06-02
Explicit bin ranges after viewport filtering

d-callan:
looking mostly for high-level feedback before i start getting into things like documentation and testing. does this seem conceptually right? are we happy to eventually do histo and lineplot binning this way as well (rather than using a viewport there)?


d-callan:
should be considered in combination w veupathdb/plot.data#163

asizemore:
Using a bin range makes sense to me, and i can see why we'll need something general to handle histos and lines when the x axis is binned and we have a discrete gradient colormap (as @d-callan you pointed out in #163 ). 

I'm still a bit behind conceptually - i'm not totally sure what using bin ranges gives us over using multiple viewports? Or is the idea that we're realizing that this concept of a bin range is more general than the viewport?

d-callan:
I guess I just see it as conceptually cleaner.. the info you need to consistently make reproducible bins is the range you're binning over and the bin width and that's all neatly in one BinSpec now. A viewport is a different but related concept I suppose in that it often determines the bin range, but I think the discretized colormap is a good ex of when the two aren't necessarily the same thing.

asizemore:
Love it - makes sense to me! Thank you!

bobular (approved):
Looks good if you want to merge now.

I commented in another ticket about my approval of moving towards a more general `binSpec` https://github.com/VEuPathDB/plot.data/issues/163#issuecomment-1147841461
Probably should have made that here!

R/class-plotdata-map-markers.R
```diff
@@ -48,12 +50,31 @@ newMapMarkersPD <- function(.dt = data.table::data.table(),
...
 
+  if (is.null(geolocationViewport)) {
+    geolocationViewport <- findGeolocationViewport(.pd, lat, lon)
+    veupathUtils::logWithTime('Determined default viewport.', verbose)
+  } else {
+    geolocationViewport <- validateGeolocationViewport(geolocationViewport, verbose)
+  }
+  if (is.null(geolocationViewport)) {
+    attr$viewport <- list('latitude'=list('xMin'=NA,
+                                          'xMax'=NA),
```
asizemore, line 61: Do we need to unbox NAs?

R/class-plotdata-map-markers.R
```diff
@@ -64,8 +85,12 @@ newMapMarkersPD <- function(.dt = data.table::data.table(),
       if (length(rankedValues) > 8) {
         rankedValues <- c(rankedValues[1:7], 'Other')
         .pd[[x]][!.pd[[x]] %in% rankedValues] <- 'Other'
+        overlayValues <- c(stringi::stri_sort(rankedValues[1:7], numeric=TRUE), 'Other')
+      } else {
+        overlayValues <- stringi::stri_sort(rankedValues[1:7], numeric=TRUE)
```
asizemore, line 90: Perhaps just `stringi::stri_sort(rankedValues, numeric=TRUE)` here? We know that the length of rankedValues is 8 or less, so we should be able to just sort the vector i'd think?

R/class-plotdata-map-markers.R
```diff
@@ -86,9 +111,17 @@ newMapMarkersPD <- function(.dt = data.table::data.table(),
...
       attr$binSlider <- binSlider
 
-      xRange <- findViewport(.pd[[x]], xType)
-      .pd[[x]] <- as.character(bin(.pd[[x]], binWidth, xRange))
+      if (is.null(binRange)) {
+        xRange <- findViewport(.pd[[x]], xType)
+      } else {
+        xRange <- validateBinRange(binRange, xType, verbose)
+      }
+      .pd[[x]] <- bin(.pd[[x]], binWidth, xRange, stringsAsFactors=TRUE)
```
asizemore, line 119: i may have missed this in a slack discussion, but is there a reason we have numBins defined before taking into account the binRange?

asizemore: Seems like it could differ from the actual total number of bins returned

R/utils-bin.R
```diff
@@ -1,20 +1,24 @@
...
 #' 
 #'
 #' This function divides the range of ‘x’ into intervals and codes 
 #' the values in ‘x’ according to which interval they fall
 #' @param x Numeric or Date vector to bin
 #' @param binWidth number to increment bin bounds by, or string for dates ex: 'month'
-#' @return Character vector of coded values 
+#' @param viewport List of min and max values to consider as the range of data
+#' @param stringsAsFactors logical indicating whether a factor or character vector should be returned
+#' @return Character vector (or factor) of coded values
```
asizemore, line 10: Out of curiosity, when would we want factors instead of strings? So far we've kept factors out of the repo. Are you foreseeing a use for them soon?

asizemore: Ohh because of this line? `overlayValues <- levels(.pd[[x]])`
And because we want to know about bins that may not have any data in them. That's why we can't do `unique(.pd[[x]])`. Okay cool :)

tests/testthat/test-utils.R
```diff
@@ -415,31 +415,63 @@ test_that("bin() returns appropriate bins for dates", {
...
+
+  dt$dateBin <- bin(as.Date(dt$entity.dateA), 'day', viewport, stringsAsFactors=T)
+  expect_equal(uniqueN(dt$dateBin), 393)
+  expect_equal(uniqueN(levels(dt$dateBin)), 3986)
+
+  viewport <- list('xMin'=min(dt$entity.contA), 'xMax'=max(dt$entity.contA))
+  binWidth <- plot.data::findBinWidth(dt$entity.contA) 
+  dt$contBin <- bin(dt$entity.contA, binWidth, viewport, stringsAsFactors=T)
+  expect_equal(uniqueN(dt$contBin), 16)
+  expect_equal(uniqueN(levels(dt$contBin)), 17)
```
asizemore, line 455: why do we get an extra bin at the upper end here? The max of `dt$entity.contA` is 14.36 which is the same as the viewport max, but the extra bin starts at 14.4.

asizemore (approved):
Left a lot of questions to help me better understand while you're away!

tests/testthat/test-utils.R
d-callan: I'd guess same reason Bob was asking about integer bin labels for non integer bin widths.. the last bin doesn't actually start at 14.4, that's just a rounded label. Haven't confirmed yet though

bobular (approved):
I've given it the once-over and it looks good to me. At least the new overlayValues logic looks good.

d-callan:
:crossed_fingers:

PR #172, opened 2022-06-02
add option to remove NAs when calculating numBinsToBinWidth

### LINKED ISSUE #169: `map-markers-overlay` sometimes fails to assign binWidth automatically (and throws 500)
My diagnosis so far

./R/class-plotdata-map-markers.R is the only code to call `numBinsToBinWidth` (to get the binWidth for 8 default bins)

I suspect it's not handling NAs correctly because in SCORE Seasonal, the Age variable causes this* unless Age is filtered on Age:0-100 (which gets rid of NAs).  Nearly all the continuous vars in SCORE Mozambique also have this problem, and can be resolved by broadly filtering on them first.

(* I'm talking about making a request to map-markers-overlay with an empty `config.binSpec` object in the request - in order to get the default 8 bins)

asizemore:
Resolves #169 

@bobular was right! The NAs were causing issues with `numBinsToBinWidth`.

Currently we use this function on the xAxisVar in map markers only. We use it _before_ creating the plot.data class, so we have to handle NAs when this function is run. Because it's an axis var and this function is run so early, it made sense to simply remove the NAs.

Still, most of our functions intentionally do not remove NAs by default, hence the default value of the new `na.rm` arg is FALSE.

I expect that we'll need to use `numBinsToBinWidth` more substantially when we implement discrete gradient colormaps in all vizs. Handling binning with overlay vars is more complicated because of the no data toggle. If we turn 'no data' on for strata vars and chose a discrete gradient colormap with a continuous overlay var that included NAs, would we want to simply remove NAs? Seems likely but worth answering independently... or just not allowing evilMode at all...

Actually maybe we should just talk about this in the data viz meeting. If we have a discrete colormap created from a continuous var with NAs, we can keep doing a no data toggle in scatter bc we're changing the marker, but not in box. In box we'd end up with, say, 8 boxes colored from the discrete colormap, and one no data box in gray. The gray no data box would not be visually distinct enough from the other 8 boxes to all individuals.

### DIFF
--- a/R/class-plotdata-map-markers.R
+++ b/R/class-plotdata-map-markers.R
@@ -73,9 +73,9 @@ newMapMarkersPD <- function(.dt = data.table::data.table(),
         } else {
           numericBinWidth <- as.numeric(gsub("[^0-9.-]", "", binWidth))
           if (is.na(numericBinWidth)) { numericBinWidth <- 1 }
-            unit <- veupathUtils::trim(gsub("^[[:digit:]].", "", binWidth))
-            binSpec <- list('type'=jsonlite::unbox('binWidth'), 'value'=jsonlite::unbox(numericBinWidth), 'units'=jsonlite::unbox(unit))
-          }
+          unit <- veupathUtils::trim(gsub("^[[:digit:]].", "", binWidth))
+          binSpec <- list('type'=jsonlite::unbox('binWidth'), 'value'=jsonlite::unbox(numericBinWidth), 'units'=jsonlite::unbox(unit))
+        }
       } else {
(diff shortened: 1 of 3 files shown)


bobular (approved):
The `na.rm` code looks good, and though I didn't go through the tests line by line, I did run them with and without `na.rm=TRUE` and it passed/failed as expected.

bobular:
Yes to talking about discrete gradient colormaps + no-data handling at dataviz!

asizemore:
Upgraded qa rserve to v3.0.3. As long as all the containers are building, @bobular you should see the update soon!

PR #176, opened 2022-06-13
improve bin rounding logic

### LINKED ISSUE #175: Map: returning extra bins
Score mozambique, var= School age children treated in village (%)

Screen Shot 2022-06-13 at 4 19 07 PM

@bobular noticed that we're returning an extra bin that's empty when the map overlay var is continuous. Why does this happen?

asizemore:
Resolves #175 

Not sure if this is truly the correct fix just yet... need to test it with a backend

Here's my idea so far:
Looks like the culprit is`cut_width`, a function that was mostly taken from ggplot. When the binWidth cuts the binRange exactly, we get an extra bin. This was added in ggplot's original version of [cut_width](https://github.com/tidyverse/ggplot2/blob/775f1b49353cf1f9354e96e3bb70155e9a56af6a/R/utilities-break.r#L81-L83). However, this workaround doesn't make sense when the binWidth cuts the binRange exactly, hence my fix. 

My suspicion is that this is what's happening with the map, but i want to test the backend to be sure.

### DIFF
--- a/R/utils-cut.R
+++ b/R/utils-cut.R
@@ -46,9 +46,20 @@ cut_width <- function(x, width, center = NULL, boundary = NULL, closed = c("righ
   min_x <- find_origin(x_range, width, boundary)
   # Small correction factor so that we don't get an extra bin when, for
   # example, origin = 0, max(x) = 20, width = 10.
-  max_x <- max(x, na.rm = TRUE) + (1 - 1e-08) * width
+  # what's happening is that seq(0, 20, 10) = c(0,10, 20)
+  # if we do the ggplot fix we'll get c(0, 10, 20, 29.9)
+  # I think we only want that when the binWidth does not divide the range perfectly.
+  max_x <- max(x, na.rm = TRUE) # + (1 - 1e-08) * binWidth
 
-  breaks <- c(seq(min_x, max_x, width), max_x)
+  breaks <- c(seq(min_x, max_x, width))
+  if (max_x %ni% breaks) { 
+    print('I do not divide evenly')
+    max_x_with_cushion <- max_x + (1 - 1e-08) * width
+    breaks <- c(c(seq(min_x, max_x_with_cushion, width)), max_x_with_cushion)
+  } else {
+    print('i divide evenly')
+  }
+  print(breaks)
   cut(x, breaks, include.lowest = TRUE, right = (closed == "right"), ...)
 }
 
--- a/R/utils-cut.R
+++ b/R/utils-cut.R
@@ -44,20 +44,25 @@ cut_width <- function(x, width, center = NULL, boundary = NULL, closed = c("righ
 
   # Determine bins
   min_x <- find_origin(x_range, width, boundary)
-  max_x <- max(x, na.rm = TRUE) # + (1 - 1e-08) * binWidth
+  max_x <- max(x, na.rm = TRUE)
 
   breaks <- c(seq(min_x, max_x, width))
   # Round breaks *before* they go into the cut function. This way the data (not rounded)
   # will be correctly divided into the rounded bins
-  breaks <- as.numeric(formatC(0 + breaks, digits = 3, width = 1L))
+  if (all(x %% 1 == 0)) {
+    avgDigits <- 0
+  } else {
+    avgDigits <- floor(mean(stringi::stri_count_regex(as.character(x), "[[:digit:]]")))
+  }
+  breaks <- as.numeric(formatC(0 + breaks, digits = avgDigits, width = 1L))
 
   # If now, after rounded, our max bin does not include the max of the data, add another bin
   if (max(breaks) < max_x) {
-    endBin <- max(x, na.rm = TRUE) + (1 - 1e-08) * width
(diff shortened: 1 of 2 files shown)


asizemore:
actually i don't think this is the right fix. The issue is that 375.24 is going into the bin that ends with 375. It's totally reasonable for 375.24 to be in the bin that is [375, 428ish]. That last bin is a good bin it just shouldn't be empty. We can see it's in the wrong bin if we filter to only that point and then see which bin is returned in rankedValues.

So, the issue is why does 375.24 get put into the wrong bin? Going to have some dinner and see if anymore lightbulbs happen

asizemore:
Can reproduce with test data. The issue is in the base R function `cut()`. It's rounding at an inconvenient time. The fix will require us to make our own version of `cut()`. 

[Here](https://stackoverflow.com/questions/37899503/inconsistent-behaviour-of-cut-different-intervals-with-same-number-and-same-d) is a very brief description

asizemore:
histogram doesn't "seem" to have this issue because it's only returning bins with data in them. The map is returning all bins that it could have, regardless of if they're empty. But in fact histogram has the same issue.

R/utils-cut.R
```diff
@@ -44,12 +44,25 @@ cut_width <- function(x, width, center = NULL, boundary = NULL, closed = c("righ
...
+  # Round breaks *before* they go into the cut function. This way the data (not rounded)
+  # will be correctly divided into the rounded bins
+  if (all(x %% 1 == 0)) {
+    avgDigits <- 0
+  } else {
+    avgDigits <- floor(mean(stringi::stri_count_regex(as.character(x), "[[:digit:]]")))
+  }
+  breaks <- as.numeric(formatC(0 + breaks, digits = avgDigits, width = 1L))
+
+  # If now, after rounded, our max bin does not include the max of the data, add another bin
```
bobular, line 59: Is this just for safety?

Is modifying the final break a viable alternative that keeps the same number of desired bins (but the last one not officially the same width, so problematic, I guess)?

```
  if (max(breaks) < max_x) {
    breaks[length(breaks)] <- max_x
  }
```

bobular (approved):
Looks great - I haven't done any practical tests, but happy to do so via the client tomorrow :-)

R/utils-cut.R
asizemore: oooh that's an interesting idea. Because we're rounding breakpoints before cutting we're already making the bins slightly different widths 😅 The fear would be that `cut` does it's own rounding. We round the bin labels before sending to `cut` to try to force it to not round. If it suddenly sees one breakpoint with a bunch of decimals and everything else with only 2 decimals, then it might round that last breakpoint again anyways.

But what we could do instead is something like 
```
if (max(breaks) < max_x) {
  breaks[length(breaks)] <- breaks[length(breaks)]  + 10^(-avgDigits)
}
```

I think if max(breaks) < max_x, then max_x must have more than `avgDigits` decimals. If it had fewer or the same, then the first rounding would have rounded the value to itself and `max(breaks) < max_x` would be false. 
`max_x` having more than `avgDigits` digits means that `max_x - max(breaks) < 10^(-avgDigits).

asizemore: i'm going to grab a snack and see if the above still makes sense after eating...

asizemore: so the math checks out, but i've since learned that even though we specify the digits in the arguments to `cut`, it's just using it as a guideline and will still do its own rounding. 
A proper fix of rewriting `cut` would take some time, so not something we can do for b58. I'll make a new issue for it so that we don't lose it. Bin labels and how they relate to the precise binWidth is a good data viz topic.

I'd vote to merge what we already have, we'll fix `cut` for b59, and also move to `numBins` for the map for b59.

bobular: looks good!

I'd have done it with `while` rather than work out if the maths is correct, but it's dangerous!

```
while (max(breaks) < max_x) {
  breaks[length(breaks)] <- breaks[length(breaks)]  + 10^(-avgDigits)
}
```

bobular: I know that's a bit sloppy.
Did you consider

```
if (max(breaks) < max_x) {
  breaks[length(breaks)] <- max_x + 10^(-avgDigits)
}
```
?

asizemore: ahh yes i did consider `max_x + 10^(-avgDigits)`, but then we would have the last bin not rounded like all the others

bobular: Ah, yes, them's the breaks.

R/utils-cut.R
```diff
@@ -44,12 +44,25 @@ cut_width <- function(x, width, center = NULL, boundary = NULL, closed = c("righ
...
+  if (max(breaks) < max_x) {
```
bobular, line 60: I think this might need to be `<= max_x` ?
I'm seeing some extra empty bins now via the EDA.

This is SCORE Mozambique. The extra bin is empty, and no data seems to be missing from the preceding 8 bins. I'll see if I can change this and run the tests. I'm not sure I can make a new test though.

image

PR #177, opened 2022-06-13
relax stop to warning for num and denom vals not in data

### LINKED ISSUE #157: 500 from lineplot when a proportion value is unknowingly filtered out
GEMS1
Line plot
x-axis: Age group
y-axis: Case or Control  case/case+control
overlay: **Lost earnings due to care**

or Borrowing, or is seems any variable uner Current care expenses.

image

What does work?

As above but with (three scenarios listed below)

- y-axis: numeric variable, like Age.
- y-axis: Participant->Sex female/male+female
- overlay: Participant->Sex

It's not obvious to me what the pattern is!
I couldn't get a similar 500 in other visualizations using the **Lost earnings due to care** overlay

asizemore:
Resolves #157 

Recall that plot.data only sees the data after it's been subsetted. So, when we ask for a proportion with num A and denom A+B, but B has been filtered out because of subsetting or otherwise, plot.data doesn't see B at all in the data. 

So, instead of erring when the client asks for the proportion with denom = A+B, the code now prints a warning that says we did not see B in the data. 

I want to think of this solution as a temporary fix. As discussed in the EDA UX meeting today, it's very unclear to a user why they will get all 1s in the case/control example shown in the issue. Instead, we should warn the user that the intersection of values has removed an entire value from the plot. I'm not sure which service should/could do that yet. But it seems possible that whatever service does it (maybe subsetting?) could just return a flag that says oops you've subsetted out an entire variable value. Later, the warning shown to the user could involve a link to some visualization that shows the interaction of values between the variables chosen so that the user will know exactly what happened.

## Testing
- so far only added one test to ensure the the line plot doesn't fail. Before the fix it just erred, so i want to ensure we get a real object out of it.
- I'd like to integration test this before merging - will do that tomorrow!

### DIFF
--- a/R/class-plotdata-line.R
+++ b/R/class-plotdata-line.R
@@ -72,7 +72,8 @@ newLinePD <- function(.dt = data.table::data.table(),
       denominatorValues <- unique(.pd[[y]])
     }
     #validate num and denom values actually exist as part of the y values
-    validateValues(numeratorValues, .pd[[y]])
+    # Removing validation for numerator as part of fixing #175. If the numerator is not in the data, we get all 0s.
+    # validateValues(numeratorValues, .pd[[y]])
     validateValues(denominatorValues, .pd[[y]])
     veupathUtils::logWithTime('Numerator and denominator values have been validated.', verbose)
 
--- a/R/utils.R
+++ b/R/utils.R
@@ -2,7 +2,10 @@
 
 validateValues <- function(valuesOfInterest, valuesOfVariable) {
   if (any(valuesOfInterest %ni% valuesOfVariable)) {
-    stop("Values of interest do not exist as real values of the specified variable.")
+    warning("At least one value of interest does not exist as a real value in the specified variable.")
+  }
+  if (all(valuesOfInterest %ni% valuesOfVariable)) {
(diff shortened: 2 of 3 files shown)


bobular (approved):
Looks good.  Do we need to test that all the values in seriesY are 1?

bobular:
> Looks good. Do we need to test that all the values in seriesY are 1?

Maybe they aren't all 1 (due to missing values or something).
Not worth doing if at all complicated!

asizemore:
That's a good idea! And not difficult to implement. Actually I started implementing and realized that this solution means that the plot will work (give all 1s) if we pick the value that's in the data, but will still err if we pick the value that's not in the data. In the example above with choosing A/A+B, this will still give you a plot if you filter away the Bs because there is data for A in the num and denom. But if we choose B/A+B we'll get an error. 

Hmm i guess it shouldn't err and instead return an empty plot?

asizemore:
okay proposal is for the numerator to be lax and the denominator to be strict. If we give the plot a numerator that is not found in the data, we get all 0s (see new test). If we give the plot denominator values, none of which are found in the data, then we err. I don't want to assume anything about dividing by 0 on accident

PR #186, opened 2022-07-21
disable automatic setting binwidth to 0 when uniqueN less than numbins

### LINKED ISSUE #185: Computed bin width = 0 for number variables
Many variables in recently reloaded datasets now have a computed bin width = 0 . Spot checked a production server (clinepidb.org  - eda058n) and the bin width there seems fine, so this must be a recent change in veupathUtils. During loading we run a script with:

`singularity exec docker://veupathdb/rserve
`
During execution a warning is printed but the process does not fail

```
1: In veupathUtils::nonZeroRound(binWidth, avgDigits) :
  Input is already zero and cannot be rounded to a non-zero number.
2: In veupathUtils::nonZeroRound(binWidth, avgDigits) :
  Input is already zero and cannot be rounded to a non-zero number.
3: In veupathUtils::nonZeroRound(binWidth, avgDigits) :
  Input is already zero and cannot be rounded to a non-zero number.

```

asizemore:
Resolves #185 

Removed automatic setting binWidth to 0 with the number of distinct values is less than the number of calculated bins.
Only commented out in case there was a reason for having this check that we missed!


asizemore:
Okay _now_ i feel better. Here are a few things I learned:
- The `bin` function was handling receiving a `binWidth=0` by assigning `bins` to be just the input array. So if i ran `bin(c(1,2,3), binWidth=0)`, then i'd get back `c(1,2,3)`. This caused tests to fail in map markers and in test-utils.
- Many of the failing tests were actually because of the above point and were giving us a false sense of security. For example in line 32 of `tests-utils.R`, we thought we were getting back 500 bins but in fact we were just returning the original array, which is why it looked like that test was happily passing. 
- We have to be careful in testing the number of bins coming out of the binning functions because we are only counting bins that actually have data. Deep in the binning functions we can find all the levels computed because we're given the bins as factors, but then we usually turn those into arrays so we only keep those bins with at least one value in them. I left a few comments about this in the testing file.
- This fix solves some problems but definitely not all our binning woes. Seems like dates are particularly difficult but not terrible, and we often get back 1 more or 1 less bin than we request. Best left for another ticket though :)

asizemore:
@bobular all ready for your review again!

bobular (approved):
Thanks, that looks great.

jaycolin:
After reloading one affected dataset, this is looking great, but there is one case not fixed. The number of distinct values = 1

https://qa.restricted.clinepidb.org/ce.restricted.qa/app/workspace/analyses/DS_3b4a4e8f09/new/variables/EUPATH_0000609/EUPATH_0021213

asizemore:
@bobular when there's one distinct value, what do you think about `bin_width = max(1, abs(rangeMax))`? That way we should just get a little bin around the one value, but that's it.

asizemore:
Sorry i actually mean to suggest `bin_width = min(1, abs(rangeMax))`

PR #187, opened 2022-08-02
Fix issue 182 where histogram bins are corrupted

bobular:
There are a couple of new tests in test-temp-tac.R that I will move to test-histogram.R before merging.

The new bin break rounding strategy is as follows:

1. Make sure at the beginning that the right-most break is >= the maximum binnable value
2. Then do the rounding increasing the number digits until the variance of the rounded binWidths is below a threshold
3. Then adjust the start and end breaks if necessary with a fancy significant digit adjustment, which will never drastically change the bin size (*no extra bins made any more at this stage*)

Note that I haven't paid any special attention to integer variables.


tests/testthat/test-temp-tac.R
```diff
@@ -0,0 +1,42 @@
...
+#
+
+test_that("histogram.dt does not produce corrupted bins when given TAC data from issue 182", {
+  map <- data.frame('id' = c('values'),
+                    'plotRef' = c('xAxisVariable'),
+                    'dataType' = c('NUMBER'),
+                    'dataShape' = c('CONTINUOUS'),
+                    stringsAsFactors = FALSE)
+
+  dt <- histogram.dt(data.issue182, map, binWidth = 1, value = 'count', barmode = 'stack', binReportValue = 'binWidth')
```
asizemore, line 12: is there a reasonable way we can modify the existing test object (like in line 25) to fail in the same way? Ideally we've been trying to keep the test data to a minimum set that can be used or manipulated for nearly all the tests. Or if we could pick out 500 values of the tac data and add them as a new column `contE` or something like that in testDF that would similarly prevent test data object sprawl, if possible.

asizemore (commented):
Thank you for finding and fixing all of these issues! I like the new approach!

Another question - if we are always creating possibly 1 or 2 more bins than expected, is that going to get us in trouble with map overlays? So sometimes we'd return 9 or 10 bins instead of 8. Is that okay?

R/utils-cut.R
```diff
@@ -50,19 +50,57 @@ cut_width <- function(x, width, center = NULL, boundary = NULL, closed = c("righ
...
+  # but first catch the case where there is only one break
+  if (length(breaks) == 1) {
+    breaks = c(breaks, tail(breaks,1) + width)
+    #I've left these debugging statements in temporarily in case you're curious
+    #print("ONLY ONE BREAK")
+  }
+
+  # note that we work with the already-rounded breaks here
+  # (we used to reround them again after adding new breaks)
+  if (min_x < head(breaks,1)) {
```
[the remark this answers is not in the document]
bobular: Just to be symmetric with the use of `tail()` that follows. Because R doesn't allow indexing from the end like `breaks[-1]`

R/utils-cut.R
```diff
@@ -50,19 +50,57 @@ cut_width <- function(x, width, center = NULL, boundary = NULL, closed = c("righ
...
+    # But the binWidth might not be an integer? Why don't we round the breaks?
+    # Should that be all(breaks %% 1 == 0) ?? (All tests pass if we do that.)
     requiredDigits <- -1
   } else {
-    requiredDigits <- ceiling(median(stringi::stri_count_regex(as.character(x), "[[:digit:]]")))
+    # Determine the appropriate number of digits to round the bin start/ends to.
+    # Start at 4 and work up!
+    requiredDigits <- 4
+    rawBreaks <- breaks
+    repeat {
```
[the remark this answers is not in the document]
bobular: I'd have preferred a `do { ... } until (COND)` loop, but it seems more loop-like than recursion to me.

bobular:
> Thank you for finding and fixing all of these issues! I like the new approach!
> 
> Another question - if we are always creating possibly 1 or 2 more bins than expected, is that going to get us in trouble with map overlays? So sometimes we'd return 9 or 10 bins instead of 8 I suppose. Is there any reason we should be concerned about this? I can't think of any but curious what you think

I think it will cause confusion. We can easily make 8, 9 or 10 colours from the continuous palette, that's no problem, but users will wonder why it changes all the time.

tests/testthat/test-temp-tac.R
bobular: We'd need to make sure that the new test data column contE fails the new test with the old code. That's making my head explode at the moment...

I suppose we could make a new branch from `main`, add the data the new way as `testDF.contE`, and copy over the new test, and make sure it fails.  Then merge that branch into this one.  Yeah, that should work.

R/utils.R
```diff
@@ -407,4 +407,58 @@ findColNamesByPredicate <- function(variableList, predicateFunction) {
...
   floor(mean(stringi::stri_count_regex(as.character(x), "[[:digit:]]")))
+}
+
+
+#
+# For any number, return an absolute delta (numeric) at the last
+# significant digit in the number, using the number of digits specified
+#
+# e.g. assuming 3 significant digits
+#
```
asizemore, line 418: We should add tests in test-utils for all of these example.

R/utils-cut.R
```diff
@@ -43,26 +43,107 @@ cut_width <- function(x, width, center = NULL, boundary = NULL, closed = c("righ
...
+  # If so, we will adjust the first and last bins by subtracting/adding one "least significant digit"
+  # from them.
+
+## DEBUG ONLY
+unmodifiedBreaks <- breaks
+
+  if (min_x < breaks[1]) {
+    newStart <- breaks[1] - signifDigitDelta(breaks[1], requiredDigits)
+    breaks[1] <- as.numeric(formatC(newStart, digits = requiredDigits, width = 1L))
+    print("MODIFIED START BIN")
```
asizemore, line 126: found it! let's use `veupathUtils::logWithTime('blah', verbose)` and i think just passing in `verbose` as an arg with default to TRUE should do it!

asizemore (approved):
Thanks for all of your thinking on this @bobular ! Looks awesome.

I'd consider moving the `signifDigitDelta` function to veupathUtils. Seems useful in more cases than plot.data i think?

R/utils-cut.R
bobular: Cool - but I don't think these tiny modifications are worth logging.  I pushed the changes in a hurry so you could take a look. I don't think these should stay in the final push.

R/utils.R
bobular: Good idea will do.

Is this more of an Epsilon than Delta?

asizemore: oh yes you're right!

tests/testthat/test-bin.R
```diff
@@ -103,6 +103,7 @@ test_that("bin() adds an extra bin only when necessary due to rounding.", {
   expect_equal(length(levels(bins)), 7)
 
   # Force rounding down at the last bin so that the max data point needs to go in an added extra bin
+  # New rounding strategy also creates a new first bin
```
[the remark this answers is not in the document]
bobular: Actually I'm surprised this test is returning 8 bins.
I designed it to return 7...
So will investigate

R/utils.R
bobular: Thanks again for your review and approval (pending all the suggestions you made)
I will aim to get this rolled out to the dfalke-b back end on Monday where I can check more easily that the bins look OK.

R/utils.R
```diff
@@ -407,4 +407,58 @@ findColNamesByPredicate <- function(variableList, predicateFunction) {
...
+  # walk through string, looking for first non-zero, non decimal point character
+  for (c in unlist(characters)) {
+    if (!(c %in% c('0', '.'))) {
+      seenSignificant <- TRUE
+    }
+    if (c == '.') {
+      result <- c(result, c)
+    } else if (seenSignificant) {
+      significantCount <- significantCount + 1
+      if (significantCount == digits) {
```
bobular, line 448: I'll put the `significantCount < digits` condition first, before `== digits` so that it is easier to understand what's going on.

R/utils-cut.R
```diff
@@ -50,19 +50,57 @@ cut_width <- function(x, width, center = NULL, boundary = NULL, closed = c("righ
...
+  if (min_x < head(breaks,1)) {
+    breaks = c(head(breaks,1) - diff(head(breaks,2)), breaks)
+    #print("EXTRA EXTRA START BIN")
   }
+
+  if (max_x > tail(breaks,1)) {
+    breaks = c(breaks, tail(breaks,1) + diff(tail(breaks,2)))
+    #print("EXTRA END BIN")
+  }
+  # sometimes, due to rounding, one extra bin isn't enough
```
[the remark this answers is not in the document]
bobular: Doesn't apply any more - not adding new bins at all!

tests/testthat/test-temp-tac.R
bobular: I've made a [new ticket](https://github.com/VEuPathDB/plot.data/issues/188) for this and the function move to veupathUtils.

R/utils.R
bobular: Not moving it at this point (see https://github.com/VEuPathDB/plot.data/issues/188) but renamed and added tests :heavy_check_mark:

tests/testthat/test-bin.R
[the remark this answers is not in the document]
bobular: I investigated and I found that the `breaks <- seq(min_x, max_x, width)` (with a calculated width that should create 7 bins) was creating a `max(breaks)` just shy of `max_x`.  By something in the 1e-16 ballpark!  So I made yet another tweak to adjust the final break to be equal to `max_x` in this case.

Reckon we should always get 8 bins for the map viz now!

R/utils.R
bobular: Done

PR #191, opened 2022-08-19
add space after comma in bins

### LINKED ISSUE #174: Legend text formatting for continuous overlays needs styling fix
Legend text formatting for continuous overlay variables is confusing. for example, `[89,657]` looks like 89.6K instead of 89 - 657

image

Can we change the formatting please? Perhaps to: `[89 to 657]`

asizemore:
Resolves #174 

Adds a space after the comma for bin labels. Not much else to say :) 

I was concerned about if commas show up in the number (14,234 for example) what would happen, but after some testing i learned we don't get commas in the numbers (which i think is good anyways), so that's not an issue.

Added to the numeric method of `bin` only because dates don't have any commas in their bin labels!


bobular (approved):
Looks good!

PR #193, opened 2022-08-24
fix completeCases equals 0 for collections  when no strata selected

### LINKED ISSUE #192: Ranked abundance: Brief message about no plottable data when selecting overlay
Only happens on ranked abundance *box plot*

To reproduce, open the ranked abundance box plot. A plot appears. Select a strata variable. A brief message about no plottable data appears (see my super fast screenshot of the message)

Screen Shot 2022-07-15 at 10 49 26 AM

It only happens if we're going from no strata vars to having one strata var. I briefly looked into this a while ago and I want to say it was `outputSize` in `BoxplotVisualization.tsx` that briefly becomes 0 which causes the error? 

I recommend waiting for VEuPathDB/web-eda#1244 to be merged before addressing this issue

asizemore:
Resolves #192 

When there were no strata vars chosen and the collection var took up both the x and y axes, we were returning `completeCasesAllVars = 0`, which caused the brief error message described in the issue. We ended up with `completeCasesAllVars = 0` because with only the collection var selected (so no non-collection vars selected), `nonCollectionVarDataRows` was empty, causing `collectionVarDataRows*nonCollectionVarDataRows` to be empty, and eventually we ended up with seeing 0 complete cases. 

## Testing
- Added additional test in `test-box.R` that uses only the collection variable (no non-collection vars)

### DIFF
--- a/R/class-plotdata.R
+++ b/R/class-plotdata.R
@@ -162,14 +162,15 @@ newPlotdata <- function(.dt = data.table(),
     }
 
     # Calculate complete cases *before* reshaping the data
+    collectionVarColNames <- veupathUtils::toColNameOrNull(collectionVariable)
     # To count complete cases, we want to only count rows where we have at least one value in one of the collection var columns
-    completeCasesPerCollectionCol <- lapply(veupathUtils::toColNameOrNull(collectionVariable), function(collectionVar) {return(complete.cases(.dt[, ..collectionVar]))})
+    completeCasesPerCollectionCol <- lapply(collectionVarColNames, function(collectionVar) {return(complete.cases(.dt[, ..collectionVar]))})
     collectionVarDataRows <- Reduce("+", completeCasesPerCollectionCol) > 0  # Any row with val=0 means that row was missing for all vars in the collection and should be removed from the count
     # Columns not corresponding to a collection var are treated differently. Calculate their complete cases as normal
-    nonColectionVarColNames <- setdiff(c(x,y,z,group, panel), veupathUtils::toColNameOrNull(collectionVariable))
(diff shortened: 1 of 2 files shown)


R/class-plotdata.R
```diff
@@ -167,9 +167,16 @@ newPlotdata <- function(.dt = data.table(),
     collectionVarDataRows <- Reduce("+", completeCasesPerCollectionCol) > 0  # Any row with val=0 means that row was missing for all vars in the collection and should be removed from the count
     # Columns not corresponding to a collection var are treated differently. Calculate their complete cases as normal
     nonColectionVarColNames <- setdiff(c(x,y,z,group, panel), veupathUtils::toColNameOrNull(collectionVariable))
-    nonCollectionVarDataRows <- complete.cases(.dt[, ..nonColectionVarColNames])
-    # Count the rows that we keep from the collection var complete cases *and* non-colection var complete cases
-    completeCasesAllVars <- jsonlite::unbox(nrow(.dt[collectionVarDataRows*nonCollectionVarDataRows,]))
+    if (length(nonColectionVarColNames) > 0) {
```
bobular, line 170: Spelling typo in `nonColectionVarColNames` (needs a double 'l')

bobular (approved):
I understand and approve the new logic.

It seems that `veupathUtils::toColNameOrNull(collectionVariable)` is calculated a few times so could be factored out.

R/class-plotdata.R
asizemore: Thank you! Collection, correlation, column, are all terribly difficult for me to spell 😄

PR #195, opened 2022-08-24
Fix194 lineplot timezone

### LINKED ISSUE #194: Bug: lineplot proportion throws client-side error
To create the bug:
1. Create a new Line Plot viz in any study (`Amazonia ICEMR Brazil Cohort` is my go-to)
2. Select an x-variable related to Dates, like `Participant: Observation Date` in Amazonia Brazil
3. Select any categorical y-variable, like `Sex` in Amazonia Brazil
4. Set the numerator for the proportion
5. Client side error will result as shown in the screenshot below

In the screenshot, I've included a glimpse of the API response from the data service.

image

asizemore:
Resolves #194 

The time zone was added a long time ago and I'm wary of a total remove this late in the game, so I added a boolean argument to `findBinStart` and `findBinEnd` that allow us to decide if we should add the time zone or not. This way, if we find out that we really only want to remove the time zone for line plot but keep it for histogram, for example, that should be a quick fix!


R/utils.R
```diff
@@ -295,14 +295,14 @@ pruneViewportAdjustmentFromBins <- function(bins, xVP, x, viewport) {
...
-findBinStart <- function(x) {
+findBinStart <- function(x, addTimeZone = c(FALSE, TRUE)) {
   if (!length(x)) {
      return(character(0))
   }
 
   if (all(grepl(" - ",x))) {
     x <- veupathUtils::strSplit(x, " - ")
-    x <- paste0(x,'T00:00:00')
+    if (addTimeZone) x <- paste0(x,'T00:00:00')
```
bobular, line 305: In the split below (and for `findBinEnd`)

the split pattern would be safer as `",\\s*"` - due to the recent post-processing of numeric bin range labels to include a space

bobular (approved):
Looks good! See my comment about splitting on comma. I don't know if this would have deleterious effects on the client, but better safe than sorry!

asizemore:
definitely better safe than sorry! And you're right, we would have been returning " 123" instead of "123" so thank you for the eagle eye review!

Issue #205, opened 2023-02-04
Visualizations: categorical overlay, find counts for customisable N<=7 + other


d-callan:
see https://github.com/VEuPathDB/EdaDataService/issues/238 for details

d-callan:
Important: very likely we dont need this for SAM so much as the map viz in EDA. this ticket is NOT a prioirity for b63.

d-callan:
since the floating charts will need this for SAM, im going to generalize this ticket to cover all plot classes and include it as a b63 priority

danicahelb:
backend ticket, nothing to QA

PR #212, opened 2023-03-14
Custom categorical overlays

d-callan:
ive assumed that overlay values are required for high cardinality categorical or continuous overlays. ive also assumed that continuous overlay values will be handled in a separate pr and that the overlay values present are from the vocabulary (meaning if i get a value like 'Other' it means i should look for that as a real value).

for categorical overlays, all values which arent among those specified by the user become 'All un-selected values', but that can be changed if needed.


d-callan:
resolves #205

R/class-plotdata-bar.R
```diff
@@ -2,6 +2,7 @@ newBarPD <- function(.dt = data.table::data.table(),
                          variables = veupathUtils::VariableMetadataList(),
                          value = character(),
                          barmode = character(),
+                         overlayValues = character(),
```
bobular, line 5: Bit confused - shouldn't `overlayValues` be passed to the `newPlotdata` a few lines below? Ah yes, I see it is in, for example, boxplot. Just an oversight!  Or a test! :smile:

R/utils.R
```diff
@@ -1,3 +1,20 @@
+recodeOverlayValues <- function(overlayData, desiredOverlayValues, variables) {
+  if (is.null(desiredOverlayValues)) return(overlayData)
+  dataType <- veupathUtils::findDataTypesFromPlotRef(variables, 'overlay')
+  
+  if (dataType %in% c('NUMBER', 'INTEGER', 'DATE')) {
+    # figure this out when i get to the continuous overlay PR, just leaving a skeleton for now
+    # tbh desiredOverlayValues will probably stop being a simple character vector then as well. but one thing at a time.
+    warning("Binned continuous overlays are not supported yet.")
+  } else {
+    if (unique(overlayData) == desiredOverlayValues) return(overlayData)
```
bobular, line 10: Does the order of `unique(overlayData)` have to be the same as `desiredOverlayValues`?

If so, something like `sort(unique(overlayData)) == sort(desiredOverlayValues)` ??

But note my lack of R prowess!

R/utils.R
```diff
@@ -1,3 +1,20 @@
...
+    overlayData[!overlayData %in% desiredOverlayValues] <- 'All un-selected values'
```
bobular, line 11: `Unselected` just one word I think?

I think the verbiage here might need some tweaking.  Imagine receiving a screenshot or share link where you don't know that the legend values were "selected".  Maybe return `__UNSELECTED__` and we figure it out client side?

bobular (commented):
Looks good - I hope my comments are useful!

R/utils.R
d-callan: I should be using all() and %in%

R/class-plotdata-bar.R
```diff
@@ -76,6 +78,7 @@ validateBarPD <- function(.bar, verbose) {
 #' @param variables veupathUtils VariableMetadataList
 #' @param value String indicating how to calculate y-values ('identity', 'count', 'proportion')
 #' @param barmode String indicating if bars should be grouped or stacked ('group', 'stack')
+#' @param overlayValues character vector providing overlay values of interest
```
asizemore, line 81: Consider adding that these values must exist in the overlay variable or else x, y , z bad things happen

R/class-plotdata.R
```diff
@@ -40,6 +41,16 @@ newPlotdata <- function(.dt = data.table(),
   lon <- veupathUtils::findColNamesFromPlotRef(variables, 'longitude')
 
   isEvil <- ifelse(evilMode %in% c('allVariables', 'strataVariables'), TRUE, FALSE)
+  collectionVarMetadata <- veupathUtils::findCollectionVariableMetadata(variables)
+  isOverlayCollection <- ifelse(is.null(collectionVarMetadata), FALSE, ifelse(collectionVarMetadata@plotReference@value %in% c('overlay'), TRUE, FALSE))
```
asizemore, line 45: is there an R power user reason for checking `value %in% c('overlay')` as opposed to `identical(value, 'overlay')`?

R/utils.R
```diff
@@ -1,3 +1,19 @@
+recodeValues <- function(values, desiredValues, dataType) {
```
asizemore, line 1: This function name for the categorical overlay didn't make sense to me. I expected swapping values for desiredValues like in a derived var. Maybe something like "filterValues" or "simplifyOverlayValues" would be a little more clear?

R/class-plotdata.R
```diff
@@ -40,6 +41,16 @@ newPlotdata <- function(.dt = data.table(),
...
+
+  if (!is.null(group) && !isOverlayCollection) {
+    groupNeedsOverlayValues <- length(data.table::uniqueN(.dt[[group]])) > 8 && useGradientColorscale == FALSE
+    if (is.null(overlayValues) && groupNeedsOverlayValues) {
+      stop("Must provide overlay values of interest for high cardinality or continuous overlay variables.")
+    }
```
asizemore, line 51: add a check to ensure all overlayValues are actually values of the varaible?

R/class-plotdata-map-markers.R
```diff
@@ -29,6 +27,13 @@ newMapMarkersPD <- function(.dt = data.table::data.table(),
   lat <- veupathUtils::findColNamesFromPlotRef(variables, 'latitude')
   lon <- veupathUtils::findColNamesFromPlotRef(variables, 'longitude')
 
+  # for the others this happen in the parent class, but color is different here
```
asizemore, line 30: why is this different? Other than we've assigned the colors to the x axis?

R/class-plotdata-bar.R
d-callan: so i didnt bc its entirely possible we could be passed valid overlay values that we wont actually see in the subset..

R/utils.R
d-callan: so some values get recoded to __Unselected__ , and for continuous vars i think they will become strings (bin labels) where they were numbers and dates. so its a bit hard to name well. but ill think about it

R/class-plotdata.R
d-callan: so i was thinking its possible we have a vocabulary like x,y,z and we request all values bc there are fewer than 8 or something, but in this particular subset we dont see all three values. so i think any type of validation would be hard.

d-callan: its possible the data service could try to validate before passing work to R.. like compare the values its passed to the known referenceMetadata

R/class-plotdata-map-markers.R
d-callan: just that i think, but  i didnt see that there was an obviously correct and cost effective way to sort that out

bobular:
Hi @d-callan - thanks for reminding me about the regular EDA map viz. The back end will no longer do 7+Other by default for high cardinality categoricals. At a minimum (to maintain status quo) the client will need to request the "value metadata" (values and their counts) and then send the top 7 with the map-markers-overlay request. At a maximum we'd provide some "paint by variable config" like in the standalone map. Am I getting this right?

d-callan:
there was a ux meeting where people said theyd be ok to just say you can only use variables w 8 or fewer values in the map, like the other vizs, and figure out high cardinality vars etc more generally at a later date. i suspect that will mean allowing some amount of user configuration for them at some point yes, but the backend is now in a better position to support that as well.

bobular (approved):
Still LGTM :-)

R/class-plotdata.R
asizemore: yeah i didn't think about that the subset might not have the desired values. Instead of a validation, maybe it'd be enough to just log any desired values that didn't appear in dt. For a non-eda user, this would be good info, and in case we have any trouble in the eda it'd be easier to see what happened in the logs.

R/class-plotdata-map-markers.R
asizemore: cool. I have no helpful suggestions, i just wanted to be sure i understood :)

PR #228, opened 2023-07-24
collection floaters

### LINKED ISSUE #227: Map: add support for collections to more backend plot classes
There is a small amount of validation that the various plot.data plot classes do for collections in box, scatter, etc. But not all plot classes do this yet. We should rectify that, and add unit tests as needed. These tests should now also include categorical collections.

d-callan:
resolves #227


R/class-plotdata-beeswarm.R
```diff
@@ -15,6 +15,7 @@ newBeeswarmPD <- function(.dt = data.table::data.table(),
                      overlayValues = overlayValues,
                      sampleSizes = sampleSizes,
                      completeCases = completeCases,
+                     inferredVarAxis = 'y',
```
asizemore, line 18: gentle beeswarm, someday maybe you will see the light of day.

R/class-plotdata-mosaic.R
```diff
@@ -134,14 +135,22 @@ mosaic.dt <- function(data, variables,
...
+  if (is.null(collectionVM)) {
+    xVM <- veupathUtils::findVariableMetadataFromPlotRef(variables, 'xAxis')
+    if (is.null(xVM)) {
+      stop("Must provide x-axis variable for plot type mosaic.")
+    }
+    yVM <- veupathUtils::findVariableMetadataFromPlotRef(variables, 'yAxis')
+    if (is.null(yVM)) {
+      stop("Must provide y-axis variable for plot type mosaic.")
+    }
   }
```
asizemore, line 148: Could merge the next conditional with an `else` here

R/class-plotdata.R
```diff
@@ -166,11 +174,6 @@ newPlotdata <- function(.dt = data.table(),
     .dt[[variable.name]] <- gsub(paste0(inferredVarMetadata@variableSpec@entityId, '.'), '', .dt[[variable.name]])
```
asizemore, line 174: i don't remember the specifics, but i do remember having a thought about how `melt`ing did or did not reorder the collection vars when they got melted. Anyways, worth eyeballing that the order that comes out of the melt is what you'd like!

asizemore (approved):
Could use some more tests. Otherwise looks great! I'm honestly impressed with how small a change this was

PR #231, opened 2023-10-13
add network and bipartite network classes

### LINKED ISSUE #229: Add Network class
For correlation app. Updated 10.13

Intakes data that looks like
```
[{
  data1: string,
  data2: string,
  correlationCoef: number,
  pvalue?: number, // ignore for now - feature upgrade for later
  adjustedPValue?: number, // ignore for now - feature upgrade for later
}] 
```

and return nodes and links, where links look like
```
[{
  source: node
  target: node
  weightData?: number
  opacityData?: number
  color: number
}]
```
nodes look like
```
[(
  id: string
 }]
```
and then we have two arrays of nodes `column1NodeIDs` and `column2NodeIDs` that let us know which nodes belong to which column.

We'd have the data service correlation bipartitie network plugin tell plot.data how to make this. So like
```
outWhatever <- bpnetwork(data=data, sourceNodes = 'var1', targetNodes = 'var2', weightData= 'correlationCoef', edgeWeightingScheme = "pos/neg", edgeThreshold=0.9)
```

asizemore:
Resolves #229 
Child issue of #234 

This PR adds two new classes: `network` and `bipartiteNetwork`. Networks do not fit the statistical chart structure so they have their own base class `network`. As we add additional networks, we can further extend this class. 

Important - this PR is meant to _only_ add the appropriate classes. Will stack PRs for the methods but i want us to focus on getting the classes right first.

To dos for this PR:

- [x] decide what empty props should be, null or character or something
- [x] network class
- [x] validation functions

Intentionally left for later PRs
- real network methods
- add helpers for making random networks for testing
- getting data from whatever format into a network
- writing to json


R/class-network-bipartite.R
```diff
@@ -0,0 +1,98 @@
+newBipartiteNetwork <- function(df = data.frame(),
+                                sourceNodeColumn = character(),
+                                targetNodeColumn = character(),
```
asizemore, line 3: is there anything from the plot.data class that we'd want here? We could make even a more abstracted class like stats chart

d-callan:
just going to comment here that we also talked about moving this to S4 right out the gate, and introducing an edgeList class so we leave clear room to grow into alternative data sources and structures. i think ill wait for those changes before reviewing this one.

R/class-Node.R
```diff
@@ -0,0 +1,88 @@
...
+#' @slot color string or numeric that determines the color of the node. Optional.
+#' @slot weight numeric value associated with the node, such as timestamp or other node-associated data. Optional.
+#' 
+#' @name Node-class
+#' @rdname Node-class
+#' @export
+Node <- setClass("Node", 
+  representation(
+    id = "character",
+    color = "ANY",
```
d-callan, line 33: I'm not sure how we're planning to manage colors yet, it's unclear how you mean this to be used.

R/class-network.R
```diff
@@ -0,0 +1,48 @@
...
+
+#' Network
+#' 
+#' A class for representing networks. A network is composed of nodes and links (edges, connections, etc.). A link is represented
+#' as a pair of nodes, with optional attributes such as weight (see Link). To represent a network, we need both the list of links in the network and a list of nodes
+#' in case some nodes have no links. A network can also have properties such as directedness, levels, colors, etc. (coming soon).
+#' 
+#' @slot links LinkList object defining the links in the network.
+#' @slot nodes NodeList object defining the nodes in the network. Some nodes may not have any links.
+#' @slot linkColorScheme string defining the type of coloring scheme the links follow. Options are 'none' (default) and 'posneg'.
```
d-callan, line 30: Posneg ? Is this at risk of being specific to correlation? What other options might we see here eventually, for clarity? Do we need a node color scheme?

R/class-Node.R
```diff
@@ -0,0 +1,88 @@
...
+    weight = "ANY"
```
d-callan, line 34: Remind me what's weight here? Truly anything? Do we need a general metadata slot? Should it take a named list maybe?

R/methods-Links.R
```diff
@@ -0,0 +1,32 @@
+# Methods for Link and LinkList objects
+
+# Accessors for fanciness
```
d-callan, line 3: I kinda like get* and set*..

R/methods-Links.R
```diff
@@ -0,0 +1,32 @@
...
+setMethod("target<-", "Link", function(object, value) {object@target <- value; validObject(object); object})
+setMethod("weight", "Link", function(object) object@weight)
+setMethod("weight<-", "Link", function(object, value) {object@weight <- value; validObject(object); object})
+setMethod("color", "Link", function(object) object@color)
+setMethod("color<-", "Link", function(object, value) {object@color <- value; validObject(object); object})
+
+
+# For LinkLists, let's return vectors of data from the nodes
+setGeneric("getSourceNodes", function(object) standardGeneric("getSourceNodes"))
+setMethod("getSourceNodes", "LinkList", function(object) lapply(object, function(x) source(x)))
```
d-callan, line 25: Could be a question here whether this would be expected to be a list of nodes vs a nodeList

R/methods-network.R
```diff
@@ -0,0 +1,82 @@
...
+
+
+# Fancy accessors
+setGeneric("getNodes", function(object) standardGeneric("getNodes"))
+setMethod("getNodes", "Network", function(object) object@nodes)
+setGeneric("getLinks", function(object) standardGeneric("getLinks"))
+setMethod("getLinks", "Network", function(object) object@links)
+setGeneric("getLinkColorScheme", function(object) standardGeneric("getLinkColorScheme"))
+setMethod("getLinkColorScheme", "Network", function(object) object@linkColorScheme)
+# No setters! Once created, a network should only be updated via network methods
```
d-callan, line 11: Even changing the color scheme?

R/methods-network.R
```diff
@@ -0,0 +1,82 @@
...
+
+# Remove isolated nodes
+# Get isolated nodes
+# Remove redundant links
+# Remove redundant nodes
+# Get Degree list
+# Get Weighted Degree list
+# etc.
+# Threshold network by edge weight
+# Assign color scheme
```
d-callan, line 21: Are these the anticipated methods?

R/class-BipartiteNetwork.R
```diff
@@ -0,0 +1,50 @@
...
+#' 
+#' @name BipartiteNetwork-class
+#' @rdname BipartiteNetwork-class
+#' @include class-network.R
+#' @export 
+BipartiteNetwork <- setClass("BipartiteNetwork", 
+  contains = "Network",
+  representation(
+    column1NodeIDs = "character",
+    column2NodeIDs = "character"
```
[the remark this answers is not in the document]
asizemore: Good question. Because I think it's simpler from a data organization perspective to view the groups/columns as properties of the network. Like this thing is "a network with these extra properties". This organization is true to concept and also easy to expand upon when we add longitudinal and multilayer networks.

R/class-network.R
```diff
@@ -0,0 +1,48 @@
+# Network class
+check_network <- function(object) {
+
+  errors <- character()
+
+  # Check that all nodes in links are in nodes
+  nodesInLinks <- NodeList(union(getSourceNodes(object@links), getTargetNodes(object@links))) # may become a method later if i find i use it elsewhere
```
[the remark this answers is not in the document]
asizemore: i have to think about the list of node lists. I can see how it'd be nice for some of the organization but my first reaction is that it will be harder to work with and no truer to the abstract concept than having properties.

R/class-Node.R
asizemore: Weight needs to be numeric (if there's not a check for that i'll go add it), but generally nodes don't have weights so i like having the default of NULL. Links have a default weight because we think of even unweighted networks as having 0s (no link) or 1s (a link) between node pairs.
I suppose setting the weight to _NA_numeric would be cleaner...

R/methods-network.R
asizemore: yes. Cannot change the color scheme without going through a network method. Otherwise i could just say setColorScheme to brown and the attribute would change but the node colors won't change. I'd rather the function be real `colorNodes(network, colorScheme)` and `colorLInks(network, colorScheme)` functions so it's more clear they are really doing a lot, not just setting a property.

R/class-network.R
asizemore: Posneg would also be good for other networks, like differential networks that have +/- some change value associated with each link. But some other options we will likely use are sequential, diverging, or avg degree of parent nodes. And yes node color schemes would be implemented the same way when we need them. Likely node color schemes are by degree, by weighted degree, by centrality, by yourFavoriteNetworkMetric, by layer

R/methods-network.R
asizemore: Oof yes. These are my plans. I had to list them to help me better organize the classes

R/class-BipartiteNetwork.R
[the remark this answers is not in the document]
asizemore: maybe i'll make it a multipartite class instead of bipartite... That would mean we have something like a partitions slot that we can add lists of ids to. 
I still like having the nodes all together because the bipartite is more a partitioning with rules of a thing than a gluing together multiple things concept.

R/class-Node.R
d-callan: I'm wondering if this isn't an inherent property of the network but maybe instead a method param? Maybe the network can just hold any relevant annotations/ data about the nodes and the method can worry about color scheme etc

R/class-network.R
d-callan: See my other comments about making color part of method params maybe

R/class-Node.R
d-callan: Still lost sry. How is the network visually impacted by this weight?

R/methods-network.R
d-callan: I think this moves me further into color as a method param territory

R/class-Node.R
d-callan: I'm pretty sure we need a way for these networks to carry arbitrary data/annotations about nodes and maybe links too

R/class-BipartiteNetwork.R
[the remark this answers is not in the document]
d-callan: Ok. I'm better able to get behind the partition slot I think and the more abstract multipartite concept.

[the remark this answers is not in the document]
d-callan: Can you say more about longitudinal and multilayer? Just for my benefit..

R/class-network.R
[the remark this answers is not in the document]
d-callan: I think Id like to hear more about longitudinal and multilayer, for a sense where we might go, before forming an opinion on the partition slot vs list of nodeLists..

R/class-BipartiteNetwork.R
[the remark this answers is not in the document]
asizemore: For multilayer and longitudinal, we can think of the layer or timepoint as a property of the node and/or link. For multilayer, i'd also assign the nodes based on `layers` slot which would be a list of node id vectors. For temporal networks, it'll depend on the data but there's a more temporal version of multilayer networks we could take or make it more general and allow links to appear and disappear as well. In the latter case, we'd timestamp the nodes and links individually, and expand the slots to allow for capturing when a node or link is present.

R/class-Node.R
asizemore: i think we're on the same page and that's where i'm going with this idea. The `linkColorScheme` slot is just there to let someone who gets the network know how the colors were mapped. Like if i said here map these values to colors you'd want to know should it be an ordinal, categorical, diverging, etc. color map to use later. This slot says "hey btw i got colored according to this principle, so use that info to render me appropriately.

R/class-Node.R
asizemore: > How is the network visually impacted by this weight?

Weight is the data associated with the node or edge that we should use when calculating network/node/link metrics (think weighted degree, modularity, etc). So for correlation, the link weight would be the correlation coeff. Think of it as the primary numeric data describing the node or edge. Weight can, but does not have to be, used to color the node or edge. It can be used directly like if we colored the bipartite network edges by correlation coeff, or indirectly like just using the sign of the correlation coeff to color edges. 

> I'm pretty sure we need a way for these networks to carry arbitrary data/annotations about nodes and maybe links too

Yes i think you're right, which is why i have nodes and links set up this way as their own classes. Basically the color, weight, etc is all just data about the node/link, but it's common enough in networks that we might as well just call it what it is. For the arbitrary data, we'd have a slot called something like "additionalData" that would be a named list. Additional data could be just more values for the object, like adding pvalue and adjusted pvalue to each link, or could be vectors or even data frames, like if each node is a time point and the data attached to it is the state of the system at that point in time. The latter options are less likely in the near term but totally possible. Anyways, i think that's how this framework would grow to accomodate extra data, and additionally i think that can be implemented later when we are really seeing the pvalues and such.

R/class-Link.R
```diff
@@ -0,0 +1,87 @@
+
+
+check_link <- function(object) {
+
+  errors <- character()
+
+  # Link color must be a string or number
+  if (!is.null(object@color) & !is.character(object@color) & !is.numeric(object@color)) {
+    errors <- c(errors, "Link color must be a string or number")
```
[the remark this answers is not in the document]
asizemore: updated messaging in recent commit

R/methods-Links.R
asizemore: is there a particular reason? If it's a preference i'd rather stick to what we have because it's more in line with what i'm guessing is more standard R practice since it follows Hadley's book.

R/methods-Links.R
asizemore: updated. I'm leaving as a list of nodes for now because i can't imagine we always really want a NodeList. But maybe in the next few weeks of using it more i'll change my mind.

R/class-BipartiteNetwork.R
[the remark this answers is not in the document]
d-callan: I think maybe what I want is to understand the difference between a partition and a layer and whether these things can coexist

R/methods-Links.R
d-callan: It's a preference, and not a particularly strong one at that. I've done it myself both ways before, but I've started recently to feel get* and set* is a bit more readable is all.

R/class-Node.R
d-callan: I need to think about this for a min

R/class-KPartiteNetwork.R
```diff
@@ -0,0 +1,47 @@
...
+#' @slot partitions list of node ids that belong to each partition
+#' 
+#' @name KPartiteNetwork-class
+#' @rdname KPartiteNetwork-class
+#' @include class-network.R
+#' @export 
+KPartiteNetwork <- setClass("KPartiteNetwork", 
+  contains = "Network",
+  representation(
+    partitions = "list"
```
d-callan, line 39: This is a list of lists of strings right? Maybe we can make it a bit more formal?

R/class-Node.R
d-callan: Maybe all we need is to add at a high level a way for the network to keep track of what varspecs map to what network attribute.

R/class-Link.R
```diff
@@ -0,0 +1,87 @@
+
+
+check_link <- function(object) {
+
+  errors <- character()
+
+  # Link color must be a string or number
+  if (!is.null(object@color) & !is.character(object@color) & !is.numeric(object@color)) {
+    errors <- c(errors, "Link color must be a string or number that represents a color or can be mapped to a color.")
```
asizemore, line 9: wait a second, why not date, too? Food for future thought. Actually that'd probably be handeled in whatever setLinkColors method accepts date data.

R/class-KPartiteNetwork.R
```diff
@@ -0,0 +1,117 @@
...
+#' An alias to NodeIdList
+#' 
+#' @name Partition-class
+#' @rdname Partition-class
+#' @export
+Partition <- NodeIdList
+
+#' Create Partitions
+#' 
+#' A list of Partition objects
```
asizemore, line 41: Mind adding some more docs here? I'm not following

asizemore: Just the usual inputs and outputs

R/class-KPartiteNetwork.R
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  return(new("Partitions", S4Vectors:::SimpleList(partitions)))
+}
+
+check_kpartite_network <- function(object) {
+
+  errors <- character()
+
+  # Check that all nodes are in at least one of the columns
```
asizemore, line 70: "columns" -> "partitions"

R/methods-Nodes.R
```diff
@@ -0,0 +1,40 @@
...
+setMethod("color", "Node", function(object) object@color)
+setMethod("weight", "Node", function(object) object@weight)
+setMethod("id<-", "Node", function(object, value) {object@NodeId@id <- value; validObject(object); object})
+setMethod("x<-", "Node", function(object, value) {object@x <- value; validObject(object); object})
+setMethod("y<-", "Node", function(object, value) {object@y <- value; validObject(object); object})
+setMethod("color<-", "Node", function(object, value) {object@color <- value; validObject(object); object})
+setMethod("weight<-", "Node", function(object, value) {object@weight <- value; validObject(object); object})
+
+## Methods for NodeId
+setMethod("id", "NodeId", function(object) object@value)
```
asizemore, line 26: should this method be "value", since the slot is named "value"?

R/methods-Nodes.R
```diff
@@ -0,0 +1,40 @@
...
+setMethod("getNodeIds", "NodeList", function(object) unlist(lapply(as.list(object), id)))
+setMethod("getWeights", "NodeList", function(object) unlist(lapply(as.list(object), weight)))
+setMethod("getColors", "NodeList", function(object) unlist(lapply(as.list(object), color)))
+
+## Methods for NodeIdList
+setMethod("getNodeIds", "NodeIdList", function(object) unlist(lapply(as.list(object), id)))
+
+## Methods for Partitions
+setGeneric("getAllNodeIds", function(object) standardGeneric("getAllNodeIds"))
+setMethod("getAllNodeIds", "Partitions", function(object) unlist(lapply(as.list(object), getNodeIds)))
```
asizemore, line 40: Would be helfpul to also have this `getAllNodeIds` for the k-partite network

asizemore (commented):
i have some questions, and i'm looking forward to checking out the next PR, but in general the approach seems good and i think it's an exciting start!

R/methods-Nodes.R
d-callan: maybe.. but value is a bit vague really. if i have a thing in hand and dont remember whether its a Node or NodeId, it might be nice to be able to use `id` in either case, is what i figured. i dont mind aliasing though for cases like these.

R/class-Link.R
d-callan: im gonna punt on this until we have a case in hand.

R/class-Network.R
```diff
@@ -0,0 +1,51 @@
...
+  representation(
+    links = "LinkList",
+    nodes = "NodeList",
+    linkColorScheme = "character",
+    variableMapping = "VariableMetadataList"
+  ), prototype = prototype(
+    links = LinkList(),
+    nodes = NodeList(),
+    linkColorScheme = 'none',
+    variableMapping = VariableMetadataList()
```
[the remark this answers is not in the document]
d-callan: no, but were also not using it yet. its a stub for now.

PR #235, opened 2023-11-02
Add network methods

### LINKED ISSUE #232: Add methods that allow us to create a network from a data.frame
More than just accessors, we need to add the following methods that will allow us to get data into a Network. Since we're only getting data as dfs for now, will only add that method, but the approach would expand just fine to matrices, etc.

- [x] get/remove isolated nodes
- [x] remove redundant links/nodes (shouldn't be allowed in a network at all so added to the network class)
- [x] threshold network by edge weight
- [x] assign colors to links
- [x] df to Network constructor
- [x] Network toJSON

asizemore:
Resolves #232 

Added methods for networks and versions for bipartite networks where appropriate. Required for getting data into a network.


d-callan:
what all are we trying to accomplish w this pr?

asizemore:
I think 1 and 2 fits into this PR. 3 is a can of worms that i'm not sure i'm ready to open yet. 4 also yes - this PR was originally about creating methods that help validate a network and get it into/out of good formats. So in addition to what you have, i'd say things like
- removing duplicate links. This should be an optional check because sometimes duplicate links are intentional
- removing duplicate nodes. I can't imagine us ever wanting duplicate nodes
- I think i alreayd have one for pruning isolated nodes
- we should have a converter into/out of whatever format igraph, etc, wants
- we also talked about converting into/out of matrix or df format for this PR a long while back.

That's a long list so breaking it up into a few PRs is also cool.

d-callan:
do we want to remove duplicate links and nodes? or just validate for them? the nodes at least it seems we just shouldnt allow that case, and then wed never need to remove them. for the links, i think id like to hear more about when wed intentionally include dups

asizemore:
yeah validating makes sense for nodes. 
For links we could also do validation and leave incorporating of multiple links as a growth area. The someday use case is let's make a network with genes as nodes and links based on coexpression values. Two genes may be coexpressed at the same level in two different experiments. That suggests two edges between the links, each with the same edge weight. In that situation one could argue that we'd find a way to also encode the two different experiments as metadata on the edge. That's possible but that info could get thrown away because it doesn't matter for whatever the question is. So there's a possibility we have two edges that are the same but we don't want to combine them, we want to consider them as two edges.

R/class-KPartiteNetwork.R
```diff
@@ -116,4 +116,75 @@ KPartiteNetwork <- setClass("KPartiteNetwork",
     partitions = Partitions()
   ),
   validity = check_kpartite_network
-)
\ No newline at end of file
+)
+
```
asizemore, line 120: i am late on this comment so it's fine to include later, but we should have check_kpartite_network check the links as well to make sure each link's source and target are from different partitions.

R/class-KPartiteNetwork.R
```diff
@@ -116,4 +116,75 @@ KPartiteNetwork <- setClass("KPartiteNetwork",
...
+  partitions = Partitions(),
+  linkColorScheme = 'none', 
+  variables = VariableMetadataList(), 
+  ...
+) {
+  # default to a single partition if none provided
+  if (length(partitions) == 0) {
+    partitions <- Partitions(list(NodeIdList(object)))
+  }
+  new("KPartiteNetwork", links=LinkList(object), nodes=NodeList(object), partitions=partitions, linkColorScheme=linkColorScheme, variableMapping=variables)
```
asizemore, line 176: Would be helpful to also have a method taking a Network and Partitions to create a KPartiteNetwork.

R/constructors-Node.R
```diff
@@ -0,0 +1,157 @@
+#' @include class-Node.R
+#' Create a Node Id
+#' 
+#' Because typing `NodeId(id = 'foo')` is annoying, this function is provided
+#' to make things easier. Now you can do `NodeId('foo')`
+#' 
+#' @param value string a unique identifier for the node
+#' @export 
+NodeId <- function(value) {
```
asizemore, line 9: yay i found you!

R/constructors-Node.R
```diff
@@ -0,0 +1,157 @@
...
+  new("NodeId", value = value)
+}
+
+
+#' Create a NodeIdList
+#' 
+#' @param object Object containing list of node ids
+#' @export 
+setGeneric("NodeIdList", function(object, uniqueOnly = c(TRUE, FALSE)) standardGeneric("NodeIdList"))
```
asizemore, line 18: what's the use case for having `uniqueOnly = FALSE`?

asizemore: I thout the validation of `NodeIdList` will return an error if the ids are not unique anyhow.
Oh is this saying "hey do you want me to unique-ify the ids you sent me?"

R/constructors-Node.R
```diff
@@ -0,0 +1,157 @@
...
+
+#' @export 
+setMethod("NodeIdList", "Node", function(object, uniqueOnly = c(TRUE, FALSE)) {
+  return(NodeIdList(list(object)))
+})
+
+
+#' Create a Node
+#' 
+#' @param id string a unique identifier for the node
```
asizemore, line 93: string, NodeId, etc

R/constructors-Node.R
```diff
@@ -0,0 +1,157 @@
...
+#' @return NodeList
+#' @export
+#' @examples
+#' NodeList(data.frame(source='a',target='b'))
+setGeneric("NodeList", function(object) standardGeneric("NodeList"))
+
+#' @export
+setMethod("NodeList", "data.frame", function(object = data.frame(source=character(),target=character())) {
+  if (!isValidEdgeList(object)) {
+    stop(paste(errors, collapse = '\n'))
```
asizemore, line 136: where does `errors` come from? I dont see it initialized here

R/methods-KPartiteNetwork.R
```diff
@@ -1,4 +1,54 @@
...
+setMethod(toJSONGeneric, "KPartiteNetwork", function(object, named = c(TRUE, FALSE)) {
+    named <- veupathUtils::matchArg(named)
+    tmp <- character()
+
+    nodes_json <- veupathUtils::toJSON(object@nodes, named = FALSE)
+    links_json <- veupathUtils::toJSON(object@links, named = FALSE)
+    partitions_json <- veupathUtils::toJSON(object@partitions, named = FALSE)
+    
+    # TODO this doesnt conform to the api in the data service, bc there we explicitly have a bipartite network and not a kpartite
+    # we have `columns1NodeIds` and `columns2NodeIds` instead of `partitions`. i think this is better though.
```
asizemore, line 46: to be clear, are you suggesting an api change? If yes, when do you think we could have this wired and tested? It's possible we could get this out with 67

R/methods-Links.R
```diff
@@ -32,3 +36,39 @@ setMethod("getWeights", "LinkList", function(object) unlist(lapply(object, funct
...
+        tmp <- paste0('{"link":', tmp, '}')
+    }
+
+    return(tmp)
+})
+
+#' @export
+setMethod(toJSONGeneric, signature("LinkList"), function(object, named = c(TRUE, FALSE)) {
+    named <- veupathUtils::matchArg(named) 
+    tmp <- veupathUtils::S4SimpleListToJSON(object, FALSE)
```
asizemore, line 69: does this work just fine using the simple list version? If yes do we even need the `Link` method just above?

R/methods-Network.R
```diff
@@ -0,0 +1,231 @@
...
+  } else {
+    veupathUtils::logWithTime('No isolated nodes found.', verbose)
+  }
+
+  validObject(net)
+  return(net)
+})
+
+getLinkUniqueString <- function(link) {
+  paste0(id(source(link)), id(target(link)))
```
asizemore, line 64: OOoh good idea. We should add a funky delimiter between the two so that the following two links don't get counted as duplicates:
1. Source: "A", Target: "BC"
2. Source: "AB", "Target: "C"

tests/testthat/test-nodes.R
```diff
@@ -84,7 +84,7 @@ test_that("NodeList methods work", {
 test_that("We cannot make nonsensical nodes", {
 
   expect_error(Node(id = FALSE))
-  expect_error(Node(id = 10))
+  #expect_error(Node(id = 10)) # why is this nonsensical?
```
asizemore, line 87: Originally node ids had to be strings. I think i saw in one of the constructors that numbers are allowed now? So we can remove this test

asizemore (commented):
My most pressing questions are around the api - how much do we want to change, and how much can we reasonably change before the release?

Additionally, I think we need to give the linkColorSchemes a bit more thought on how we protect someone from making something nonsensical with link colors that are any float but the color scheme says "posneg" for example. 

Finally, we added a bunch of methods but i think the coverage on testing is pretty low. Given how much testing we'd really want to add it's probably better if we make a separate PR to increase coverage. 

If we do decide to push for the api change before 67, i'm okay with making new issues for the other changes and starting those after the api change is done

R/class-Node.R
```diff
@@ -36,18 +36,6 @@ setClass("NodeId",
   validity = check_node_id
 )
 
-
-#' Create a Node Id
```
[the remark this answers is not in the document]
d-callan: i think i separated the constructors into their own file. felt like it was getting out of hand..

R/class-Node.R
```diff
@@ -196,12 +155,10 @@ check_node_list <- function(object) {
 #' @rdname NodeList-class
 #' @importFrom S4Vectors SimpleList
 #' @export
-NodeList <- setClass("NodeList", 
+setClass("NodeList",
```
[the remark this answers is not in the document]
d-callan: setClass will automatically create a generator/ constructor fxn. i dont want it, so im not assigning it to anything, and instead making my own.

R/class-Link.R
```diff
@@ -37,14 +38,49 @@ Link <- setClass("Link",
...
+setGeneric("Link", function(source, target, weight = 1, color = NULL, isDirected = FALSE) standardGeneric("Link"), signature = c("source", "target"))
+
+#' @export
+setMethod("Link", c("Node", "Node"), function(source, target, weight = 1, color = NULL, isDirected = FALSE) {
+  new("Link", source = source, target = target, weight = weight, color = color, isDirected = isDirected)
+})
+
+#' @export
+setMethod("Link", c("character", "character"), function(source, target, weight = 1, color = NULL, isDirected = FALSE) {
+  Link(source = Node(source), target = Node(target), weight = weight, color = color, isDirected = isDirected)
```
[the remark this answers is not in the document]
d-callan: yea, thanks to some clever/ custom constructors

R/class-Link.R
```diff
@@ -37,14 +38,49 @@ Link <- setClass("Link",
...
+})
+
+#' @export
+setMethod("Link", c("NodeId", "NodeId"), function(source, target, weight = 1, color = NULL, isDirected = FALSE) {
+  Link(source = Node(source), target = Node(target), weight = weight, color = color, isDirected = isDirected)
+})
+
+#' @export
+setMethod("Link", c("missing", "missing"), function(source, target, weight = 1, color = NULL, isDirected = FALSE) {
```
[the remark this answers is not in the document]
d-callan: its just so that `Link()` if called for some reason, wont err but just produce an empty Link. we got this for free when we do things like `Link <- setClass("Link")` and i just dont want to lose it despite my custom constructors.

R/methods-KPartiteNetwork.R
d-callan: yes, im suggesting an api change. the work for that would be in the data service and monorepo. and i think if we do it, it should be for b67

R/class-Link.R
```diff
@@ -83,10 +119,80 @@ check_link_list <- function(object) {
...
+
+    return(link)
+  }
+
+  linkList <- apply(object, 1, makeLink, linkColorScheme)
+  new("LinkList", linkList)
+})
+
+#' @export
+setMethod("LinkList", "missing", function(object) {
```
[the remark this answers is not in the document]
d-callan: 'missing' isnt an object, its nothing at all. this lets someone call `LinkList()` w no args and not err.

R/methods-Links.R
d-callan: i dont understand the question i think..... but yes, we need the Link method. it gets called by S4SimpleListToJson

R/class-Link.R
```diff
@@ -83,10 +119,80 @@ check_link_list <- function(object) {
...
   ),
   validity = check_link_list
-)
\ No newline at end of file
+)
+
+#' Generate a LinkList
+#' 
+#' Generate a LinkList from an edgeList
+#' @param object Object containing data to be converted to a LinkList
```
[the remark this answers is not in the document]
d-callan: i suppose so. it could be a number of different objects, which i guess i could list here.

R/class-Link.R
```diff
@@ -83,10 +119,80 @@ check_link_list <- function(object) {
...
+#' @examples
+#' LinkList(data.frame(source='a',target='b'))
+setGeneric("LinkList", function(object, linkColorScheme = c('none', 'posneg')) standardGeneric("LinkList"), signature = c("object"))
+
+#' @export
+setMethod("LinkList", "data.frame", function(object = data.frame(source=character(),target=character()), linkColorScheme = c('none', 'posneg')) {
+  if (!isValidEdgeList(object)) {
+    stop(paste(errors, collapse = '\n'))
+  }
+
```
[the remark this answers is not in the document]
d-callan: i think i need more info here. what is it youre trying to prevent? when would we need to coerce data to fit a color scheme? 

if youd like i can put out a warning that the color scheme arg will be ignored if color is already set, and not set the color scheme slot. that seems reasonable enough to start.

R/class-Network.R
```diff
@@ -14,10 +14,18 @@ check_network <- function(object) {
...
+  if (data.table::uniqueN(getNodeIds(object@nodes)) < length(getNodeIds(object@nodes))) {
+    errors <- c(errors, 'Duplicate node ids found. Node ids must be unique.')
+  }
+
 
   return(if (length(errors) == 0) TRUE else errors)
 }
 
+## TODO
+## i wonder if i can do something like `Network <- setClass("Network", slots = c(links = "LinkList", nodes = "NodeList"))`
```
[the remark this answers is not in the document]
d-callan: ill remove the comment. i already played around a bit and decided it wasnt worth pursuing.

R/constructors-Node.R
d-callan: magic! jk lol. i need to fix this up, and write some tests w bad edgeLists. isValidEdgeList should either be true or a list of errors, and i was moving too quick when i wrote this.

R/class-Link.R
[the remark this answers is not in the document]
d-callan: im starting to wonder if an edgeList should be allowed to have a color column. its completely ambiguous what the expected behavior should be in this case and i dont have a real use case for it i think. do you?

d-callan:
ill make a separate issue for test coverage, and we can make a list of things wed like to test there and decide whats a good time to do that. everything else i think i addressed today.

im also in process of updating the api here to match the data service, and write bipartitenetwork explicitly if we know we have two partitions for the kpartite network. thatll be in here as well.

R/class-Link.R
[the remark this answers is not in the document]
asizemore: I'm trying to prevent a Network from being created that has values in the link color slots that don't match the `linkColorScheme` slot. I like the warning idea, also we could just assign the `linkColorScheme` to be "custom" in the case where we're handed color when making a Network and not adding color via one of our setLinkColor (or whatever we called it) functions. 
No i don't have any near-term use cases. Eventually we might want a network with edges colored by the type of evidence used to assign that edge (ex. correlation, two-hybrid screen, etc), which would end up with us likely getting a list of link colors from somewhere else. But that's far away.

R/constructors-Node.R
asizemore: for my own benefit, how might the silent errors get produced in this case?

R/class-KPartiteNetwork.R
```diff
@@ -1,82 +1,51 @@
-#' @include methods-KPartiteNetwork.R
-check_partitions <- function(object) {
-  errors <- character()
-  
-  if (!!length(getAllNodeIds(object))) {
-    # Ensure that no node is in multiple partitions
-    if (length(getAllNodeIds(object)) > data.table::uniqueN(getAllNodeIds(object))) {
-      errors <- c(errors, 'Found a node in multiple partitions. Nodes can only exist in one partition.')
+  isLinkWithinPartition <- function(link, partitions) {
```
asizemore, line 1: Suggestion to rename to "isIntraPartitionLink", becuase the field often calls links "inter-" vs "intra-" something.

asizemore (commented):
I can't approve a PR i co-authored but if i could i would! 🎉

R/class-Link.R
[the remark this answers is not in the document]
d-callan: ive just prevented custom color schemes for now, and figure we can revisit when we meet a use case

PR #237, opened 2024-01-03
Empty plotrefs

d-callan:
in the popbio megastudy for the map, we have these fun variables w 'study dependent vocabs' which we need to use in order to know which values on one variable we need to impute zeroes for on another variable..... that means we need to pass these variables and their vocabs around various places, and they might eventually end up here w empty plot refs. plot.data can simply ignore anything w an empty plot ref, since that means its not actually in the plot its producing data for.


tests/testthat/test-box.R
```diff
@@ -1931,3 +1931,58 @@ test_that("box.dt() returns same shaped outputs for string cats and num cats.",
...
+    new("VariableMetadata",
+      variableClass = new("VariableClass", value = 'native'),
+      variableSpec = new("VariableSpec", variableId = 'cat5', entityId = 'entity'),
+      # empty plotReference should be ignored
+      dataType = new("DataType", value = 'STRING'),
+      dataShape = new("DataShape", value = 'CATEGORICAL'))
+  ))
+
+  dt <- box.dt(df, variables, 'none', FALSE, computeStats = T)
+  # dt is the right shape, *doesnt* include cat5 in col names
```
asizemore, line 1975: does it matter if dt contains the cat5 column? I think no if i remember right... i think somewhere in the plot.data class we chop off any columns that aren't part of the plot.

asizemore (approved):
nice

tests/testthat/test-box.R
d-callan: yea the parent plot.data class looks for specific plot refs and retains only variables associated w those.

PR #243, opened 2024-02-15
Network tests

### LINKED ISSUE #241: Backend: improve test coverage of network classes in plot.data
we added a lot of cool stuff [ here ](https://github.com/VEuPathDB/plot.data/pull/235) that have only very basic tests. we should make a list here of tests wed like to add and then decide when is a good time to improve test coverage.

d-callan:
resolves #241

### DIFF
--- a/R/class-Node.R
+++ b/R/class-Node.R
@@ -39,7 +39,7 @@ setClass("NodeId",
 check_node_id_list <- function(object) {
   errors <- character()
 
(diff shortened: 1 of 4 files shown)


d-callan:
i made a first pass.. what else do we think needs to be here?

asizemore (approved):
nothing jumped out at me as missing, so lgtm!

PR #244, opened 2024-02-16
Degree list

d-callan:
resolves https://github.com/VEuPathDB/plot.data/issues/239


d-callan:
i didnt do anything w weighted degree for now.

tests/testthat/test-network.R
```diff
@@ -147,13 +147,16 @@ test_that("We can remove links by weight", {
...
   )
   nodeB <- Node(
-    id = NodeId('B')
+    id = NodeId('B'),
+    degree = 2
   )
   nodeC <- Node(
-    id = NodeId('C')
+    id = NodeId('C'),
+    degree = 2
```
asizemore, line 159: i dont think the degree should allowed to be set when creating a node. The degree is not a thing outside of the network context. First, I could have nodeA in network1 and network2. There's no reason nodeA should have the same degree in both. 
Second, i could set it wrong. Why would i know that nodeA will have two links when i haven't made the edge list yet? These degrees aren't connected to the edge list at all

asizemore:
Idk where my review button went but anyways
- the degree part above is most important. Having the degree be a node prop is fine, but it should only ever be set within the context of a Network because it has the EdgeList. A lone node doesn't have a degree or any concept of what that means becuase it's just a node. Something like `net <- setNodeDegrees(net)`.

tests/testthat/test-network.R
d-callan: what if i just dont export it from the package, and it can only ever be used internally? i dont see why, if you have an edgeList in hand, you shouldnt be able to create the node already knowing the degree. but i agree random people shouldnt make random nodes w random values for degree.

d-callan: i think id even be happy if the only thing we actually did export were the edgeList based constructors, particularly for now

PR #245, opened 2024-02-16
Layouts

d-callan:
adding support for some layouts via igraph


R/constructors-Node.R
```diff
@@ -133,25 +133,44 @@ setMethod("Node", "missing", function(id, x = numeric(), y = numeric(), color =
...
+      coords <- igraph::layout_nicely(graph)
+    } else {
+      stop("layout must be 'force', 'circle' or 'nicely'")
+    }
+    rownames(coords) <- names(igraph::V(graph))
+  }
 
-  makeNodeWithDegree <- function(nodeId, allNodeIds) {
-    new("Node", id = NodeId(nodeId), degree = length(which(allNodeIds == nodeId)))
+  # if we want to move this out of the constructor it needs to have graph and coords passed to it
```
asizemore, line 165: agreed we should move it out of constructors. This is a thing dan would want to do

R/methods-Nodes.R
```diff
@@ -42,6 +42,14 @@ setMethod("getWeights", "NodeList", function(object) unlist(lapply(as.list(objec
 setMethod("getColors", "NodeList", function(object) unlist(lapply(as.list(object), color)))
 setGeneric("getDegrees", function(object) standardGeneric("getDegrees"))
 setMethod("getDegrees", "NodeList", function(object) unlist(lapply(as.list(object), degree)))
+setGeneric("getCoords", function(object) standardGeneric("getCoords"))
+setMethod("getCoords", "NodeList", function(object) {
+    coords <- data.frame(x = unlist(lapply(as.list(object), x)), y = unlist(lapply(as.list(object), y)))
```
asizemore, line 47: is it weird that the coords would come out with no nodeIds? I go back and forth

asizemore (commented):
I don't remember - Node already has x and y slots, right?

R/class-Network.R
```diff
@@ -100,10 +103,12 @@ setMethod("Network", signature("data.frame", "missing", "missing"), function(
   links, 
   nodes,
   linkColorScheme = 'none', 
+  layout = c("nicely", "force", "circle"),
```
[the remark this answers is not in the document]
d-callan: i was planning to leave support for other contexts for later. the first thing we need is the edgeList for the taxa X taxa app. i can file a ticket w next steps for networks layouts.

R/methods-Nodes.R
d-callan: yea. idk. igraph did it that way, but thats not necessarily a reason we should.. ill think about it.

R/class-Network.R
```diff
@@ -64,7 +64,10 @@ setClass("Network",
 #' @param links LinkList
 #' @param nodes NodeList
 #' @param object Object containing data to be converted to a Network
-#' @param linkColorScheme string defining the type of coloring scheme the links follow. Options are 'none' (default) and 'posneg'.
+#' @param linkColorScheme string defining the type of coloring scheme the links follow. 
+#' Options are 'none' (default) and 'posneg'.
+#' @param layout string defining the layout of the network. Options are 'force', 'circle',
```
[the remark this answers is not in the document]
d-callan: yea ill add this to the future network layout ticket when i make it.

R/class-Network.R
```diff
@@ -64,7 +64,10 @@ setClass("Network",
 #' @param links LinkList
 #' @param nodes NodeList
 #' @param object Object containing data to be converted to a Network
-#' @param linkColorScheme string defining the type of coloring scheme the links follow. Options are 'none' (default) and 'posneg'.
+#' @param linkColorScheme string defining the type of coloring scheme the links follow. 
+#' Options are 'none' (default) and 'posneg'.
+#' @param layout string defining the layout of the network. Options are 'force', 'circle', 
+#' and 'nicely' which are implemented in igraph. Default is 'nicely'.
```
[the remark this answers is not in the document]
d-callan: nice is a nice option though. it goes through different algorithms i dont remember based on some network metrics i dont remember. and as the default, people dont have to ask for nicely, or ever see thats what its called. they just get it by default, and can ask for specific things by name. i cant imagine ever asking for a thing called nicely without knowing what it was..

R/constructors-Node.R
d-callan: i can move it i suppose, but it seems we just as easily could later if an actual use case came up. in either case, i dont understand what it is dan would be wanting here.

R/methods-Nodes.R
d-callan: do getColors and getDegrees return named lists? i guess thats the real question.. we should be consistent.

R/constructors-Node.R
asizemore: I guess if we leave dan to use igraph then this can stay as is. I keep thinking general purpose for how this should all work but it's just the ds for now that needs to do this.

d-callan: i guess at the heart of it, the point is that being able to get a specific node w the degree and coords populated out of an igraph object is generally useful? thats interesting, i hadnt anticipated that. but i also think how to do that best for general audiences is a thing we can save for later.

R/class-Network.R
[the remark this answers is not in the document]
asizemore: i wonder how i ended up writing that suggestion three times... yikes

R/methods-Nodes.R
d-callan: those dont appear to be named in the result, so im going to leave it. if we decide we dont like it someday, we can change them all at once as a separate issue.

d-callan:
as promised https://github.com/VEuPathDB/plot.data/issues/248

PR #253, opened 2024-03-18
switch bins to closed on left

d-callan:
depends on https://github.com/VEuPathDB/veupathUtils/pull/40

resolves https://github.com/VEuPathDB/web-monorepo/issues/762


asizemore (approved):
So this was an issue only with the integers?

d-callan:
yes and no.. the integers and dates where bin width was 1 made it obvious, bc there was only a single value in the starting bin and it wasnt the value the label in the plot showed.

PR #264, opened 2025-02-07
Have scatter return point ids

### LINKED ISSUE #263: Scatter return point ids
Currently plot.data doesn't return point ids. It shouldn't have to, but given ids let's have plot.data pass them along. This is important for the scatterplot, for example, so that we can get tooltips with proper names on hover.

asizemore:
Resolves #263 

The scatterplot will now return the ids of points if an id column is given _and_ `returnPointIds = TRUE`. 

NOTE: This feature was not implemented for the case where `value=density` nor for when the `correlationMethod != 'none'`. I attempted both, found the former very tough to debug, and found that the latter already was erring (see `main`). Since neither are part of production code, nor do we have any plans to make them production code, I'm leaving them as is. I also commented out the failing correlation tests.

### DIFF
--- a/R/class-plotdata.R
+++ b/R/class-plotdata.R
@@ -48,10 +48,8 @@ newPlotdata <- function(.dt = data.table(),
   lon <- veupathUtils::findColNamesFromPlotRef(variables, 'longitude')
 
   # If we ask for the point ids, ensure the column is present. Otherwise set to null. 
-  print(returnPointIds)
-  print(idColumn)
   if (!is.null(returnPointIds) && length(idColumn) > 0) {
-    if (returnPointIds && !is.null(idColumn) && idColumn %in% names(.dt)) {
+    if (idColumn %in% names(.dt) && nrow(.dt) == uniqueN(.dt[[idColumn]])) {
       idCol <- idColumn
     } else {
       idCol <- NULL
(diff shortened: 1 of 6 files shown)


R/class-plotdata.R
```diff
@@ -45,6 +47,17 @@ newPlotdata <- function(.dt = data.table(),
   lat <- veupathUtils::findColNamesFromPlotRef(variables, 'latitude')
   lon <- veupathUtils::findColNamesFromPlotRef(variables, 'longitude')
 
+  # If we ask for the point ids, ensure the column is present. Otherwise set to null. 
+  if (!is.null(returnPointIds) && returnPointIds && length(idColumn) > 0) {
+    if (idColumn %in% names(.dt) && nrow(.dt) == uniqueN(.dt[[idColumn]])) {
+      idCol <- idColumn
+    } else {
+      idCol <- NULL
```
bobular, line 55: Hi! Should this maybe throw an error? Seems like there will be downstream trouble if we asked for point IDs and there's no column to get them from?

bobular (commented):
Looks good - just one comment!

R/class-plotdata.R
asizemore: That's a good idea! I'll add it.

bobular:
Looks good to me on my phone! Consider it approved.

